PR #4, opened 2024-01-22
feat: run clippy + BlueBuildTrait

n16hth4wkk:
- Run clippy and add clippy warnings for better code cleanness
- Condense commands to a trait and shove them into a folder together


src/commands/mod.rs
```diff
@@ -0,0 +1,15 @@
+#[cfg(feature = "init")]
+pub mod init;
+
+pub mod build;
+pub mod local;
+pub mod template;
+
+pub trait BlueBuildCommand {
+    fn run(&mut self);
```
gmpinder, line 9: I'm thinking, maybe create a default implementation that checks if try_run errors, prints the error, and exits the process. That'll help reduce more repeated code

src/lib.rs
```diff
@@ -1,16 +1,12 @@
 //! The root library for blue-build.
-
+#![warn(clippy::correctness, clippy::suspicious, clippy::perf, clippy::style)]
 #![doc(
     html_logo_url = "https://gitlab.com/wunker-bunker/blue-build/-/raw/main/logos/BlueBuild-logo.png"
 )]
 #![doc = include_str!("../README.md")]
 #![forbid(unsafe_code)]
 #![allow(unused_imports)]
+#![allow(clippy::module_name_repetitions)]
```
gmpinder, line 9: What does this end up doing?

gmpinder (changes requested):

src/lib.rs
n16hth4wkk: clippy throws warnings if the sturct or trait has the same ending as the file name.

ex: ContainerFileTemplate -> template.rs

I find this pretty annoying lol I like to defined an ending no matter the file

src/commands/mod.rs
n16hth4wkk: oh yeah this will help alot

src/lib.rs
gmpinder: Ok yeah makes sense

src/commands/build.rs
```diff
@@ -493,6 +500,9 @@ impl BuildCommand {
         Ok(image_name)
     }
 
+    /// # Errors
+    ///
+    /// Will return `Err` if the build fails.
```
gmpinder, line 505: Was this part of the pedantic mode?

gmpinder (approved):

PR #8, opened 2024-01-22
fix: Rebase path not being generated properly

gmpinder:
I move the format logic for appending the archive directory and the `oci-archive:` prefix in the `generate_full_image` function. This also fixes using the wrong string for the `rebase` command.


n16hth4wkk (approved):
lg

PR #9, opened 2024-01-23
ci: Create GitHub Workflow

n16hth4wkk:
Add support for building, tagging, and releasing via CICD


.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: |
+            target/
+            ~/.cargo/bin/
+            ~/.cargo/git/db/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+          key: ${{ runner.os }}-cargo-build-stable-${{ hashFiles('**/Cargo.toml') }}
+
```
gmpinder, line 43: You actually won't need this, the cargo build targets in earthly use a shared cache with each other. Also this won't get mounted in (unless you can find a way to do so)

.github/workflows/tag.yml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          version: v0.8.0
+
+      # Pulling rust toolchain & install linux dependencies
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Add yq (for reading Cargo.toml)
+        uses: mikefarah/yq@v4.40.5
+        with:
+          cmd: ""
```
gmpinder: `yq` comes with `tomlq`? Thats good to know

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+      - uses: actions/checkout@v3
+      - uses: chainguard-dev/actions/setup-gitsign@main
+
+      # Pulling rust toolchain & install linux dependencies
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Install cargo-release
+        uses: taiki-e/install-action@v1
+        with:
+          tool: cargo-release
```
gmpinder, line 33: Will this update the changelog? I've been using [cargo smart-release](https://github.com/Byron/cargo-smart-release)

.github/workflows/build.yml
gmpinder: Though it would be useful in the release section in case you need to do a `cargo install cargo-smart-release`

{
gmpinder, line 1: Oops?

gmpinder (changes requested):

.github/workflows/tag.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+
+      #Deps
+      - name: Set up Earthly
+        uses: earthly/actions-setup@v1
+        with:
+          use-cache: true
```
gmpinder, line 33:
```suggestion
        with:
          version: v0.8.0
          use-cache: true
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+      DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+      DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+
+    steps:
+      - uses: earthly/actions-setup@v1
+        with:
+          version: v0.8.0
+
+      # Pulling rust toolchain & install linux dependencies
+      - uses: dtolnay/rust-toolchain@stable
```
gmpinder, line 30:
```suggestion
        with:
          use-cache: true
          version: v0.8.0
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      # Setup repo and add caching
+      - uses: actions/checkout@v4
+
+      - name: Docker Login
+        run: docker login --username "$DOCKERHUB_USERNAME" --password "$DOCKERHUB_TOKEN"
```
gmpinder, line 36: Let's login to ghcr instead so we can push the images there

gmpinder (changes requested):

gmpinder:
I'm going to add some `Earthfile` changes for setting the image repo correctly

.github/workflows/tag.yml
n16hth4wkk: do we want to lock to 0.8 or is latest ?

gmpinder: Oh if we leave it out it will be latest? I think that's a good idea

Earthfile
```diff
@@ -1,12 +1,8 @@
-VERSION \
-	--global-cache \
-	--use-function-keyword \
-	--arg-scope-and-set \
-	0.7
+VERSION 0.8
 
 IMPORT gitlab.com/wunker-bunker/ci-pipelines/earthly/cargo AS cargo
```
gmpinder, line 3:
```suggestion
IMPORT github.com/blue-build/earthly-lib/cargo AS cargo
```

I'm gonna make us a new repo for earthly functions

gmpinder (changes requested):

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+          version: v0.8.0
+
+      # Pulling rust toolchain & install linux dependencies
+      - uses: dtolnay/rust-toolchain@stable
+
+      # Setup repo and add caching
+      - uses: actions/checkout@v4
+
+      - name: Docker Login
+        run: docker login ghcr.io -u $DOCKERHUB_USERNAME -p $DOCKERHUB_TOKEN
```
gmpinder, line 37:
```suggestion
        run: docker login ghcr.io -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
```

Then you can remove the `DOCKERHUB` env vars

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+      # Pulling rust toolchain & install linux dependencies
+      - uses: dtolnay/rust-toolchain@stable
+
+      - name: Install cargo-release
+        uses: taiki-e/install-action@v1
+        with:
+          tool: cargo-smart-release
+
+      - name: Release
+        run: cargo smart-release -ue
```
gmpinder, line 46: Do we need to do anything to make sure this can push updates to the repo? It will make commits and tags

.github/workflows/tag.yml
```diff
@@ -0,0 +1,61 @@
...
+      - "v*.*.*"
+
+env:
+  FORCE_COLOR: 1
+  CARGO_TERM_COLOR: always
+  NIGHTLY_TOOLCHAIN: nightly
+
+jobs:
+  tag:
+    timeout-minutes: 30
```
gmpinder, line 15:
```suggestion
    timeout-minutes: 60
```
Just in case

.github/workflows/tag.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+      - name: Add tomlq
+        run: pip install tomlq
+
+      - name: Docker Login
+        run: docker login ghcr.io -u $DOCKERHUB_USERNAME -p $DOCKERHUB_TOKEN
```
gmpinder, line 44:
```suggestion
        run: docker login ghcr.io -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
```

gmpinder (changes requested):

.github/workflows/release.yml
n16hth4wkk: i want to try a test run live to see what happens

gmpinder: Let's do it!

gmpinder (approved):

PR #10, opened 2024-01-24
doc: Update README.md

n16hth4wkk:
update description

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 # BlueBuild
 
-This project attempts to create a more concise version of ublue starting point all condensed into a single Rust based CLI tool.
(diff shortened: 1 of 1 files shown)


xynydev:
Repetion of the word concise is not the least of my problems with this. Mention of startingpoint is another one. I think this needs to be rewritten, still. Something like below could be good for now.

> BlueBuild's command line program that builds Containerfiles and custom images based on your recipe.yml.

n16hth4wkk:
> Repetion of the word concise is not the least of my problems with this. Mention of startingpoint is another one. I think this needs to be rewritten, still. Something like below could be good for now.
> 
> > BlueBuild's command line program that builds Containerfiles and custom images based on your recipe.yml.

okay im fine with that until we create better docs

gmpinder (approved):

PR #11, opened 2024-01-24
chore: use GHCR for install.sh

xynydev:
Trying to make the action install the binary from the container on GHCR with `podman run --rm ghcr.io/blue-build/cli:main-installer | sudo bash` (like in the README). Getting some error and this _might_ be related and might not, but shouldn't hurt to merge either, since I just made the `cli` package public.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,12 +6,12 @@ function cleanup() {
   echo "Cleaning up image"
   podman stop -i -t 0 blue-build-installer
   sleep 2
-  podman image rm ghcr.io/blue-build/cli:main-installer
+  podman image rm ghcr.io/blue-build/cli:latest-installer
 }
 
-podman pull ghcr.io/blue-build/cli:main-installer
+podman pull ghcr.io/blue-build/cli:latest-installer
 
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -6,12 +6,12 @@ function cleanup() {
   echo "Cleaning up image"
   podman stop -i -t 0 blue-build-installer
   sleep 2
-  podman image rm registry.gitlab.com/wunker-bunker/blue-build:latest-installer
+  podman image rm ghcr.io/blue-build/cli:main-installer
```
gmpinder, line 9: Oh you'll want that to be latest-installer. The problem is we haven't made a tag with the new GitHub workflow so it's not there yet. This tag would be following the main branch builds directly

gmpinder (changes requested):

install.sh
xynydev: Oh, yeah, saw that there's no `latest-*` on GHCR, so I just assumed main=main branch=latest. I guess latest means latest release, then?

gmpinder (approved):

xynydev:
There still isn't a container tagged as "latest"...

gmpinder:
Alright it's pushed

PR #22, opened 2024-01-27
feat: add release workflows

akarui-hikarii:


.github/workflows/release.yml
```diff
@@ -32,17 +35,42 @@ jobs:
...
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
 
       - name: Release
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: cargo smart-release -ue
+
+      # Leaving bevy comments here for us to discuss. Do we want to start tagging our pre-releases with `dev`?
```
akarui-hikarii, line 48: @gmpinder what do you think about using dev tag and then when we release, remove the dev tag

gmpinder: I'd be down to do that

.github/workflows/release.yml
```diff
@@ -32,17 +35,42 @@ jobs:
...
-          tool: cargo-smart-release
+      - name: Git setup
+        run: |
+          git config user.name github-actions
+          git config user.email github-actions@github.com
 
       - name: Release
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: cargo smart-release -ue
```
gmpinder, line 46: You'll want to delete this

.github/workflows/release.yml
```diff
@@ -18,10 +19,12 @@ jobs:
...
           fetch-depth: 0
-      - name: Git setup
-        run: |
-          git config user.name github-actions
-          git config user.email github-actions@github.com
+
+      - uses: chainguard-dev/actions/setup-gitsign@main
+
+      - name: Install cargo-release
+        run: cargo install cargo-release
```
gmpinder, line 26: You might want to move this after the cache step

.github/workflows/post-release.yml
```diff
@@ -0,0 +1,57 @@
...
+            jq --raw-output '.packages | .[] | select(.name == "blue-build").version')
+
+          # Sanity check: current version should be 0.X.Y
+          if ! grep -q '^0\.[0-9]\+\.[0-9]\+$' <<< "${current_version}"; then
+            echo "Invalid version (not in 0.X.Y format): ${current_version}"
+            exit 1
+          fi
+
+          minor_version=$(sed 's/^0\.\([0-9]\+\).*/\1/' <<< "${current_version}")
+          next_version=0.$((minor_version + 1)).0-dev
```
gmpinder, line 38: Just a note, we'll need to update this later to account for the major version too as this seems to only be concerned about versions `0.x.x`

gmpinder:
If cargo release doesn't handle publishing to crates.io, we'll need to update the tag workflow to do that or make it part of the post-release process

akarui-hikarii:
Cargo release should push to xrates. _should_

.github/workflows/post-release.yml
akarui-hikarii: noted. i think for bevy they havent release a 1.0.0 version yet so

.github/workflows/release.yml
```diff
@@ -32,17 +32,40 @@ jobs:
             ~/.cargo/registry/cache/
           key: ${{ runner.os }}-cargo-build-stable-${{ hashFiles('**/Cargo.toml') }}
 
-      - uses: chainguard-dev/actions/setup-gitsign@main
+      - name: Install cargo-release
+        run: cargo install cargo-release
 
-      # Pulling rust toolchain & install linux dependencies
-      - uses: dtolnay/rust-toolchain@stable
```
gmpinder, line 38: I think we might need this?

akarui-hikarii: i dont know. bevy doesnt have it so im curious

gmpinder (approved):

PR #37, opened 2024-02-02
fix: Add missing container tags

gmpinder:


src/module_recipe.rs
```diff
@@ -136,6 +148,33 @@ impl<'a> Recipe<'a> {
             process::exit(1);
         })
     }
+
+    fn get_os_version(&self) -> Result<String> {
+        trace!("Recipe::get_os_version()");
+        check_command_exists("skopeo")?;
```
bayou-brogrammer, line 154: What happens if this doesnt exist on the user system?

gmpinder: Then it errors and the program exits, I could just default us to using the `image_version` defined in the recipe if we really want.

bayou-brogrammer: Something to fallback on unless we force the CLI to  install the binary

bayou-brogrammer (approved):

PR #38, opened 2024-02-03
feat: Use COPY syntax for files module

gmpinder:


templates/Containerfile.module
```diff
@@ -6,6 +6,12 @@
 {{ self::print_containerfile(c) }}
 				{%- endfor %}
 			{%- endif %}
+		{%- else if type == "files" %}
```
bayou-brogrammer, line 9: It is possible to manipulate the whitespace so we don't have to do weird manual white spai

bayou-brogrammer (approved):

templates/Containerfile.module
gmpinder: The indents are actually not necessary. I just put them in to better see the logic structure. Would you prefer we didn't indent?

PR #43, opened 2024-02-04
fix: allow for image registry variables for local builds

bayou-brogrammer:
For local builds, this is needed to sign

### DIFF
--- a/templates/Containerfile
+++ b/templates/Containerfile
@@ -6,10 +6,8 @@ LABEL io.artifacthub.package.readme-url=https://raw.githubusercontent.com/ublue-
 LABEL io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
 
 ARG RECIPE={{ recipe_path.display() }}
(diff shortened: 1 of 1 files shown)


xynydev (approved):

templates/Containerfile
```diff
@@ -6,10 +6,8 @@ LABEL io.artifacthub.package.readme-url=https://raw.githubusercontent.com/ublue-
 LABEL io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
 
 ARG RECIPE={{ recipe_path.display() }}
-{%- if self::running_gitlab_actions() %}
 ARG IMAGE_REGISTRY=ghcr.io/ublue-os
 COPY cosign.pub /usr/share/ublue-os/cosign.pub
-{%- endif %}
```
gmpinder, line 12: Let's instead put this behind a check for the presence of the cosign.pub file. That way Gitlab builds can continue to work without it. (This should have been `running_github_actions` instead)

gmpinder (changes requested):

PR #45, opened 2024-02-04
feat: Allow default recipe path

gmpinder:


src/constants.rs
```diff
@@ -0,0 +1,3 @@
+pub const COSIGN_PATH: &str = "./cosign.pub";
+pub const MODULES_PATH: &str = "./config/modules";
+pub const RECIPE_PATH: &str = "./config/recipe.yaml";
```
xynydev, line 3: it's .yml by default

gmpinder:
```suggestion
pub const RECIPE_PATH: &str = "./config/recipe.yml";
```
Thanks for that!

bayou-brogrammer (approved):

PR #49, opened 2024-02-05
fix: Lowecase registry and update IMAGE_REGISTRY arg

gmpinder:


fiftydinar:
I'll approve with my limited (non-knowing) understanding of Rust code.

I trust this code.

fiftydinar (approved):

src/commands/build.rs
```diff
@@ -376,20 +376,24 @@ impl BuildCommand {
...
             ) {
                 (_, _, _, _, Some(registry), Some(registry_path)) => {
                     trace!("registry={registry}, registry_path={registry_path}");
                     format!(
                         "{}/{}/{}",
-                        registry.trim().trim_matches('/').to_lowercase(),
-                        registry_path.trim().trim_matches('/').to_lowercase(),
-                        recipe.name.trim().to_lowercase()
+                        registry.trim().trim_matches('/'),
+                        registry_path.trim().trim_matches('/'),
```
bayou-brogrammer, line 395: trim().trim_matches() looks like a code smell lol

bayou-brogrammer (approved):
this looks good im not nitpiking.

We should protentially look at something whihc just lowercases all our core varialbes we need.

src/commands/build.rs
gmpinder: The first trim removes whitespace, the second removes any `/` characters on the start and end so that we don't accidentally have a double `//` in the image path

PR #52, opened 2024-02-08
fix: remove deprecated bling `COPY` for `files` and `rpms`

gerblesh:
https://github.com/ublue-os/bling/blob/main/Containerfile the containerfile copies an empty folder in these locations for backwards compatability. `/tmp/files` and `/tmp/rpms` have been deprecated from bling, all of the RPMs have mostly moved to COPR or ublue-os/config


gmpinder:
Looks like there is a problem running builds for fork PRs. I'll have to look into this later. I'll test locally and merge if it's good

gmpinder (approved):

PR #55, opened 2024-02-09
chore: Create base integration test setup

gmpinder:
This will allow us to create a suite of tests for our new features rather than using my personal image repo


bayou-brogrammer (approved):
good skeleton structure. nothing from me atm. LGTM

gmpinder:
> good skeleton structure. nothing from me atm. LGTM

Yeah I'm hoping to add some logic to the mock scripts to we can make sure we're sending the correct args without having to run the programs themselves.

PR #57, opened 2024-02-09
fix: Rebase/Upgrade not using local image cache

gmpinder:
When building an image directly to an oci-archive, podman and buildah wouldn't use any of the cache in the local registry. This fix makes it so that it will build the image in the local registry and then export it to a local archive file.


gmpinder:
Doing some more local testing, this change isn't necessary.

PR #62, opened 2024-02-14
feat: Use GitHub's OIDC for signing images

gmpinder:
This change will follow the patter that was established with the GitLab pipeline. Users will be able to have their images signed by GitHub's OIDC. This will allow users to rebase directly onto a signed image and forego using the siging module.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -26,7 +26,7 @@ use tokio::runtime::Runtime;
 
 use crate::{
     commands::template::TemplateCommand,
-    constants::RECIPE_PATH,
+    constants::{GITHUB_TOKEN_ISSUER_URL, RECIPE_PATH},
     module_recipe::Recipe,
     ops::{self, ARCHIVE_SUFFIX},
 };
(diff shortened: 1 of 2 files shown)


src/commands/build.rs
```diff
@@ -552,6 +552,8 @@ fn sign_images(image_name: &str, tag: Option<&str>) -> Result<()> {
         env::var("SIGSTORE_ID_TOKEN"),
         env::var("GITHUB_EVENT_NAME"),
         env::var("GITHUB_REF_NAME"),
+        env::var("GITHUB_WORKFLOW_REF"),
+        env::var("GITHUB_SERVER_URL"),
```
bayou-brogrammer, line 556: does github provide these?

gmpinder: They sure do https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

gmpinder: Wait I don't even think I need the SERVER_URL

bayou-brogrammer (dismissed):

bayou-brogrammer (approved):

PR #67, opened 2024-02-17
feat: Use tmpfs mount for /tmp and /var

gmpinder:
This allows scripts to create as many files as they need in /tmp and /var without having to worry about them being included in the final image. Now the last instruction will only be the ostree container commit

### DIFF
--- a/templates/Containerfile
+++ b/templates/Containerfile
@@ -72,6 +72,8 @@ COPY {{ src }} {{ dest }}
 			{%- endif %}
 		{%- else %}
 RUN \
+	--mount=type=tmpfs,target=/tmp \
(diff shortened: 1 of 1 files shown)


bayou-brogrammer:
Should be give a final stage as a cleanup stage for users? They might not know to cleanup after and have a bloated image

gmpinder:
> Should be give a final stage as a cleanup stage for users? They might not know to cleanup after and have a bloated image

I'm not sure I follow? Users and module owners won't have to worry about cleaning up in `/tmp` or `/var` with the `tmpfs` mount. This mount is used in one build and tossed after. It's not baked into an image layer.

xynydev (approved):

PR #69, opened 2024-02-17
feat: Allow user to use source images

gmpinder:


gmpinder:
I have a plan to use multi-stage builds for this instead of overlaying all the modules into a single bind mount. This way we can allow the image name to be the key and allow the use of 3rd party modules that could have the same name and not collide.

gmpinder:
Realized my initial implementation was way more complicated than it needed to be. Instead we just use the source directly in the `/module` bind mount. This makes it so that the source will not overwrite any user defined modules and will only be used for the one module it is defined on.

bayou-brogrammer (approved):

PR #74, opened 2024-02-19
feat: block overriding

bayou-brogrammer:
This PR helps transition users who may not realize that we override their Containefile.

I added two Constants:
- `CONTAINER_FILE_PATH` => `./Containefile`

I am not worried about the tag name, we can always come up with something better, but let me know what yall think of the build date. I kind of like knowing when this build kicked off.


src/commands/template.rs
```diff
@@ -181,6 +181,14 @@ fn print_containerfile(containerfile: &str) -> String {
     file
 }
 
+#[must_use]
+fn get_build_date() -> String {
+    debug!("get_build_date()");
```
gmpinder, line 186:
```suggestion
    trace!("get_build_date()");
```

gmpinder (changes requested):

templates/Containerfile
```diff
@@ -47,6 +47,7 @@ COPY --from=ghcr.io/blue-build/cli:
 latest-installer
 {%- endif %} /out/bluebuild /usr/bin/bluebuild
 
+ARG BLUE_BUILD_DATE={{ self::get_build_date() }}
```
gmpinder, line 50: How about we turn this into a `LABEL` instead and use the constant like we do for `BUILD_ID_LABEL` on line 21.

src/ops.rs
```diff
@@ -27,6 +27,24 @@ pub fn check_command_exists(command: &str) -> Result<()> {
...
 }
 
+pub fn check_file_modified(file: &str) -> Result<bool> {
+    trace!("check_file_modified({file})");
+    debug!("Checking if {file} is modified");
+
+    // Git returns 0 if the file is not modified, 1 if it is
+    let is_dirty = Command::new("git")
+        .args(["status", "--porcelain", "|", "grep", file])
+        .output()?;
```
gmpinder, line 37: Does this actually work? I would suggest using `Stdio::piped()` to send output from one command to another. [Example](https://chat.openai.com/share/99e1658c-8ca4-47eb-a7e2-3a89608c7fca) from ChatGPT (don't use the `expect` plz).

templates/Containerfile
bayou-brogrammer: think we can just use BUILD_ID_LABEL

src/commands/template.rs
bayou-brogrammer: removed

bayou-brogrammer:
PR will fail. I know the solution locally. will fix in the AM

src/commands/build.rs
```diff
@@ -146,8 +151,34 @@ impl BlueBuildCommand for BuildCommand {
...
+                    bail!("Failed to read Containerfile: {e}")
+                }
+            };
+
+            if !file.lines().any(|line| {
+                line.to_string()
+                    .trim()
+                    .starts_with(&format!("LABEL {}", constants::BUILD_ID_LABEL))
+            }) {
+                warn!("Containerfile has not been modified by BlueBuild, skipping build");
```
xynydev, line 174: These errors need to be written to be human-friendly and self-documenting. Read a bit of the CLI interface guidellnes: https://clig.dev/#errors

Here's an example:
> It looks like you have a Containerfile in this directory that has not been generated by BlueBuild. Do you want to continue building and overwrite the Containerfile with a BlueBuild-generated one (Y, n)?

src/ops.rs
```diff
@@ -23,6 +23,33 @@ pub fn check_command_exists(command: &str) -> Result<()> {
     }
 }
 
+pub fn check_file_modified(file: &str) -> Result<bool> {
```
xynydev, line 26: I'm not sure if checking for a modification known to git in the Containerfile (if I'm interpreting this code correctly) is a good way to determine whether the file is safe to overwrite. In the case that a user is migrating from the legacy template, they'll probably have all changes commited to git..

I'd rather check only if the file is in `.gitignore` and if it has the bluebuild label.

- File in `.gitignorr`, has label -> build normally
- File in `.gitignore`, no label -> build normally
- File not in `.gitignore`, has label -> "It looks you have a BlueBuild-generated Containerfile that is not .gitignored. Do you want to remove it from git and add it to .gitignore (Y, n)?" build either way
- File not in `.gitignore`, doesn't have label -> *error message one similar to the one in my earlier message* only build on yes

xynydev: Great! Looking sweet!

templates/Containerfile
```diff
@@ -52,40 +52,40 @@ ARG IMAGE_NAME="{{ recipe.name }}"
...
 {%- for module in recipe.modules_ext.modules %}
-  {%- if let Some(type) = module.module_type %}
-    {%- if type == "containerfile" %}
-      {%- if let Some(containerfiles) = self::get_containerfile_list(module) %}
-        {%- for c in containerfiles %}
+{%- if let Some(type) = module.module_type %}
+{%- if type == "containerfile" %}
+{%- if let Some(containerfiles) = self::get_containerfile_list(module) %}
+{%- for c in containerfiles %}
 {{ self::print_containerfile(c) }}
```
gmpinder, line 59: Please stop removing the indents here. It's hard to follow the scoping without it

gmpinder (changes requested):

templates/Containerfile
bayou-brogrammer: Must be the linter or formatted because I'm not touching the file 👍

src/commands/build.rs
```diff
@@ -146,8 +160,55 @@ impl BlueBuildCommand for BuildCommand {
...
+                        } else {
+                            NO_LABEL_ERROR_MESSAGE
+                        }
+                        .color(Colors::BrightYellowFg),
+                    )
+                    .default(true)
+                    .build();
 
+                if let Ok(answer) = requestty::prompt_one(question) {
+                    if answer.as_bool().unwrap() {
```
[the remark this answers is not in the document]
bayou-brogrammer: I understand the hesitation but using unwrap in a scenario that most likely will never fail isn't really a problem. I can try to find a "cleaner" work around,

templates/Containerfile
gmpinder: Yeah the PR I'll be doing to seperate out all this stuff, I'm going to give this file the `.j2` file extension so it has some sort of proper file highlighting

src/commands/build.rs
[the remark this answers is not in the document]
gmpinder: Since it's a bool, just use `.unwrap_or_default` or `.unwrap_or` with the safer path being the "default"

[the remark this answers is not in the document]
bayou-brogrammer: i think defaulting to false would be preferred?

src/commands/build.rs
```diff
@@ -146,8 +160,55 @@ impl BlueBuildCommand for BuildCommand {
...
+        //   - If doesn't => *Build*
+        //   - If it does:
+        //     - check entry in .gitignore
+        //       -> If it is => *Build*
+        //       -> If isn't:
+        //         - check if it has the BlueBuild tag (LABEL)
+        //           -> If it does => *Ask* to add to .gitignore and remove from git
+        //           -> If it doesn't => *Ask* to continue and override the file
+
+        let container_path = format!("./{}", constants::CONTAINER_FILE);
```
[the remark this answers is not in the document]
bayou-brogrammer: oh missed that

gmpinder (approved):
Made some changes to clean up the constants, add the `--force` flag, and simplified the append_to_file function

PR #75, opened 2024-02-20
refactor: Use GITHUB_TOKEN instead of REGISTRY_TOKEN

gmpinder:
To keep things as consistent as possible, we're switching to using the GITHUB_TOKEN env var for login instead. Env vars were also all pulled out into their own constants to make things more consistent.

This change also includes prioritizing public/private key signing over OIDC keyless for GitHub for an easier transition. It would require the user to delete their `cosign.pub` file from their repo in order to start using the keyless method.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -34,12 +34,7 @@ use tokio::{
     sync::oneshot::{self, Sender},
 };
 
-use crate::{
-    commands::template::TemplateCommand,
-    constants::{GITHUB_TOKEN_ISSUER_URL, RECIPE_PATH},
-    module_recipe::Recipe,
-    ops::{self, ARCHIVE_SUFFIX},
-};
+use crate::{commands::template::TemplateCommand, constants::*, module_recipe::Recipe, ops};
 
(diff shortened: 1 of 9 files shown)


src/commands/build.rs
```diff
@@ -658,25 +651,27 @@ fn sign_images(image_name: &str, tag: Option<&str>) -> Result<()> {
...
-            debug!("On {github_ref_name} branch");
+            info!("Signing image: {image_digest}");
 
-            info!("Signing image {image_digest}");
+            trace!("cosign sign --key=env://COSIGN_PRIVATE_KEY {image_digest}");
 
-            trace!("cosign sign {image_digest}");
             if Command::new("cosign")
                 .arg("sign")
+                .arg("--key=env://COSIGN_PRIVATE_KEY")
```
gmpinder, line 674: The section here looks weird in diff form, but I basically just moved the match arm up above the keyless match arm.

bayou-brogrammer (approved):

bayou-brogrammer:
definitely like the constant var cleanup

PR #77, opened 2024-02-22
fix: allow empty custom modules dir

qoijjj:


gmpinder:
This won't work like you expect. Create an issue and I'll get a fix in for you tonight.

qoijjj:
Doing so now. Out of curiosity, why won't this work?

qoijjj:
https://github.com/blue-build/cli/issues/78

xynydev:
> Out of curiosity, why won't this work?

The COPY command copies from outside the container to inside the container. The RUN command only works inside the container. So basically, this change would create a modules directory inside the container, then copy the modules directory from outside the container to inside the container.

gmpinder (approved):

PR #80, opened 2024-02-23
refactor: Move modules into their own directory structure

gmpinder:
This separates out the module template logic so that it is easier to undertstand what each section does. This will also allow creating in repo documentation for any special modules that don't use a bash base.


bayou-brogrammer (approved):
definitely helps clean it up. noice!

xynydev (approved):

PR #82, opened 2024-02-24
docs: Add module documentation for 'containerfile' and 'files'

gmpinder:


fiftydinar:
This looks good!

Should we depreciate bash `files` module for this one?
What is the plan for that?

gmpinder:
> This looks good!
> 
> Should we depreciate bash `files` module for this one? What is the plan for that?

I think we should since we aren't using any of the bash from that module.

xynydev:
I'm not sure about including the files module docs here, because the bash module will be kept alive too, but the docs here are essentially the same.

Will review other docs more closely later today.

xynydev:
I opted to put the `files` module docs only here instead, removing the docs from the bash modules repo.

PR #83, opened 2024-02-24
refactor: Move templates to their own crate

gmpinder:
This PR logically separates out parts of the code to their own crates. As we've been adding more and more logic into the template for the Containerfile, compilation times have started to increase. This is mostly due to the fact that askama is a compilation time templating engine that uses macros to build rust code. I figured the best way to help with that was to seperate out the templating into its own crate. In order to do that while keeping the CLI as the root package, I had to also move ops, constants, and the recipe structs into their own sub-packages.


gmpinder:
I'm not planning on merging this until we've gone through the release this weekend. I just decided on a whim to do some extra organizing.

Cargo.toml
```diff
@@ -1,51 +1,80 @@
...
+description.workspace = true
+repository.workspace = true
+license.workspace = true
+
 [dependencies]
-anyhow = "1"
-askama = { version = "0.12", features = ["serde-json", "serde-yaml"] }
-chrono = "0.4"
+blue-build_utils = { path = "./utils" }
+blue-build_recipe = { path = "./recipe" }
```
bayou-brogrammer, line 44: Should these blue_build_*

The mismatch casing is odd

gmpinder:
Further testing on my end showed not a significant decrease in compilation times. However, this refactor would still be good for future development when we move towards a Tauri app.

bayou-brogrammer (approved):

Cargo.toml
gmpinder: We should keep it consistent with the root crate so how about using `blue-build-*` instead.

gmpinder: Per conversation in discord, we're going with `-`.

bayou-brogrammer (approved):

PR #85, opened 2024-02-25
feat(init): Add the new/init subcommands

gmpinder:
# Tasks

- [x] Finish setting up signing driver
- [x] Add logic for generating keys
- [x] Add logic for updating description and image name in  recipe file while preserving comments (no serde for this)
- [x] Add logic for asking the user for input
  - [x] Also add a way to use defaults and ignore questions


process/drivers/traits.rs
```diff
@@ -12,17 +12,50 @@ use semver::{Version, VersionReq};
...
+impl_private_driver!(Driver);
+impl_private_driver!(DockerDriver);
+impl_private_driver!(PodmanDriver);
+impl_private_driver!(BuildahDriver);
+impl_private_driver!(GithubDriver);
+impl_private_driver!(GitlabDriver);
+impl_private_driver!(LocalDriver);
+impl_private_driver!(CosignDriver);
+impl_private_driver!(SkopeoDriver);
+impl_private_driver!(CiDriverType);
```
gmpinder, line 51: Repeat over the type so that we can have a comma delimited list of types instead

PR #88, opened 2024-02-26
refactor: Update build command to use BuildStrategy and add `docker` support

gmpinder:


gmpinder:
Still working on the refactor, but after this it should be good to start bringing in support for docker

src/commands/build.rs
```diff
@@ -249,140 +218,27 @@ impl BlueBuildCommand for BuildCommand {
...
-        trace!("BuildCommand::build_image({client:#?})");
+        // #[cfg(feature = "builtin-podman")]
+        // match BuildStrategy::determine_strategy()? {
+        //     BuildStrategy::Socket(socket) => Runtime::new()?.block_on(self.build_image_podman_api(
+        //         Podman::unix(socket),
+        //         build_id,
+        //         &recipe_path,
+        //     )),
+        //     _ => self.build_image(&recipe_path),
+        // }
```
gmpinder, line 230:
```suggestion
```

src/strategies/build_strategy.rs
```diff
@@ -0,0 +1,413 @@
...
+};
+
+use crate::commands::build::Credentials;
+
+#[derive(Debug)]
+enum Strategy {
+    Buildah,
+    Podman,
+    #[cfg(feature = "builtin-podman")]
+    Socket(Arc<Podman>, Runtime),
```
bayou-brogrammer, line 42: This might need to include docker as well. A generic inside arc might be needed

src/strategies/build_strategy.rs
```diff
@@ -0,0 +1,413 @@
...
+                })?;
+            }
+        };
+        Ok(())
+    }
+
+    pub fn tag(&self, src_image: &str, image_name: &str, tag: &str) -> Result<()> {
+        let dest_image = format!("{image_name}:{tag}");
+
+        match &self.strategy {
```
bayou-brogrammer, line 193: So I notice we are matching alot. What if we create a trait like our clap commands and have them specify generic commands we can call?

src/strategies/build_strategy.rs
gmpinder: No that won't work in this case as this is a specific Podman Api variant with code pertaining to its library.

src/strategies/docker_strategy.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, TypedBuilder)]
+pub struct DockerStrategy {
+    creds: Option<Credentials>,
+}
+
+impl BuildStrategy for DockerStrategy {
+    fn build(&self, image: &str) -> Result<()> {
+        trace!("docker build . -t {image}");
+        let status = Command::new("docker")
+            .arg("build")
```
bayou-brogrammer, line 20: we should use buildx command instead

bayou-brogrammer: https://docs.docker.com/build/buildkit/

gmpinder: I looked into it and it and `docker build` is now the alias to `docker buildx build`.

bayou-brogrammer: I thought you had to pass DOCKER_BUILDKIT

bayou-brogrammer (dismissed):

bayou-brogrammer (approved):

PR #90, opened 2024-02-26
feat: allow for files path

bayou-brogrammer:


gmpinder (changes requested):
Add some tests in the integration tests dir to ensure this change works please.

xynydev:
What's the aim here? Is there a need to include files from beyond `config/files/`? We had some previous discussion about something similar, but I'd be interested in what sparked this PR.

bayou-brogrammer:
> What's the aim here? Is there a need to include files from beyond `config/files/`? We had some previous discussion about something similar, but I'd be interested in what sparked this PR.

This just allows for the possibility to move your files out of the config dir if you wish. 

To me this doesn't break our model of having our config dir be the source of truth. It allows for backwards comparability with the default being the config dir. 

I personally like to have the system files be on the root level of my dir instead of nested inside config

xynydev:
Backwards compatibility with what? I get the desire to have the files to include files in the root of the repo, that's what I would have done from the beginning if the earlier structure allowed for it. We now, however, have plenty other modules that require similar dirs in the config folder, so it makes more sense.

Maybe this is is a fine exception, though.

gmpinder:
I personally don't mind it being here, I just want to make sure there aren't any objections before we merge. I personally wouldn't use it.

@bayou-brogrammer Can you make sure you update the README for the module too?

bayou-brogrammer:
I don't want to introduce something that will cause strife with the developers. We are either united on this decision, or it doesn't move forward. I want this tool to be a community managed tool, so all input is valid

gmpinder:
> We now, however, have plenty other modules that require similar dirs in the config folder, so it makes more sense.

Well I guess with this mindset, we would allow a user to have this ability for all modules? Though that might be a bit harder to handle considering we would need to require module authors to be sure they are using the supplied env vars. If they did, we would be able to set `CONFIG_DIRECTORY` for the specific module being run and then the user would have to know that they would have to make sure they're still naming the directory correctly for the module to pick up the correct files.

So I guess what I'm saying is it's possible and a rather easy implementation to do, but would we want to deal with the support when users start doing weird things with their configs?

gmpinder:
Meh you know I don't really care so long as we have the documentation for the `files` module. @xynydev up to you if you have any problem. I'll stop bike shedding

bayou-brogrammer:
It doesn't seem like all our modules use the config dir. But it's easily managed to either use config dir or a custom dir. 

I'll also leave this to @xynydev 

If you prefer we stick with the config folder. I'll just use a custom container file snippets

xynydev:
I agree with the usefulness or an existence of a desire to have the files COPYed directly into the image in the root of the repository, but I'm not certain on it's inclusion or exclusion. I do not think that the other modules need to support a similar configuration, though, as their files are clearly part of their `config`.

> I don't want to introduce something that will cause strife with the developers. We are either united on this decision, or it doesn't move forward. I want this tool to be a community managed tool, so all input is valid

This is a way to express one of the reasons I am not inclined to just shut things down if I feel uncertain about them, lol.

gmpinder:
In that case if @xynydev and I are ambivalent, lazy consensus dictates this is fine. Just make sure to update the README @bayou-brogrammer and I'll approve.

bayou-brogrammer:
image

xynydev:
Docs needs more work. Adding to the default configuration in module.yml is one thing. There's also an almost duplicated paragrah I see, but I'm not sure if that:s an issue that existed previously or something that was introduced in the confusingly diffed docs commit.

I can handle the docs updates directly if yall can wait for it.

fiftydinar:
I am honestly a fan of `config/files` structure, which is standardized with other modules that utilize that structure.
Some modules don't have that structure simply because they do not require the user to input files, but yaml entries only.

Having more flexibility for `files` module may satisfy some users, but some added flexibility can have some cost:
1. It would make troubleshooting slightly harder if outside users would try to help, as they would have to look into files `path:` entry to see where those are located & then proceed to help. When you know that any recipe config resides in `config` folder, it makes it easier to troubleshoot. Configuration of the image is not just yaml, but files too.
2. It can happen that some users would do dumb things, like setting path to files in `/modules` directory, which we would want to avoid. This would be a rare edge-case, but it can still happen.

Having this flexibility omitted wouldn't mean that we limit the users in customization, but that we are simply forming standards that makes things simpler & easier to maintain in the future.

xynydev:
Convincing arguments. Another way to address these would be to only allow setting the path to the root of the repo or `config/files/`.

bayou-brogrammer:
At this point its mostly just something I want in bluebuild, but I can have a workaround with using my own module or even containerfile bling module

bayou-brogrammer:
Im going to close this for now. We can revisit if we have users asking for it

xynydev:
Just using the containerfile module is a good option.

Issue #95, opened 2024-02-27
v1.1.x breaks some images with an error about content in var


qoijjj:
```
Found file: var/lib/alternatives/ld
error: Found content in var
```

https://github.com/secureblue/secureblue/actions/runs/8068238477/job/22040524220#step:8:999

Does not occur on 1.0.2.

gmpinder:
So it seems like this directory was created outside a module. In our attempt to use tmpfs mounts for var, we didn't consider users/base images put content in there. We'll have to add back the rm for tmp and var to keep this working.

qoijjj:
@gmpinder it's not something we (secureblue) are doing since we made no changes to those images, so it's likely something happening in ucore. To be clear, it's only the ucore-nvidia images we're seeing this in:

https://github.com/ublue-os/ucore/blob/main/ucore/Containerfile

gmpinder:
> @gmpinder it's not something we (secureblue) are doing since we made no changes to those images recently, so it's likely something happening in ucore. To be clear, it's only the ucore-nvidia images we're seeing this in:
> 
> https://github.com/ublue-os/ucore/blob/main/ucore/Containerfile

Right I'm saying this is on us to fix

qoijjj:
@gmpinder I know, I was just adding some additional context :smile:

PR #103, opened 2024-02-27
fix: Run `ostree container commit` at the end of each module run

### LINKED ISSUE #101: fix: proper usage of `ostree container commit`
I was just reading the [CoreOS documentation on OSTree Native Containers](https://coreos.github.io/rpm-ostree/container/) and saw this:

> ## Using “ostree container commit”
> 
> In a container build, it’s a current best practice to invoke this at the end of each RUN instruction (or equivalent). This will verify compatibility of /var, and also clean up extraneous files in e.g. /tmp.
> 
> In the future, this command may perform more operations.

We should remove the additional `RUN ostree container commit` step, and instead add it to the end of each `RUN` command. 

This was not an issue previously with the legacy template, as all the modules were run as a single layer.

Read more: https://github.com/ostreedev/ostree-rs-ext/issues/159

gmpinder:
Since the command `ostree container commit` checks for the presence of mounted directories, we will be running it at the end of each module run. We have also updated the final commit to remove from /tmp/ and /var/ again in case a user creates extra files through custom instructions.

Closes #101 #95

### DIFF
--- a/template/templates/Containerfile.j2
+++ b/template/templates/Containerfile.j2
@@ -48,9 +48,6 @@ ARG BASE_IMAGE="{{ recipe.base_image }}"
 
 COPY --from=gcr.io/projectsigstore/cosign /ko-app/cosign /usr/bin/cosign
 COPY --from=docker.io/mikefarah/yq /usr/bin/yq /usr/bin/yq
-
-{%- include "modules/modules.j2" %}
-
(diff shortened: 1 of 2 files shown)


gmpinder:
@qoijjj I ran this against your failing build locally before and after the change. It appears to fix it.

bayou-brogrammer (approved):

PR #105, opened 2024-02-28
fix: address akmods latest bug

bayou-brogrammer:
Akmods do not support the latest tag. This just sets it to 39 if they have latest in their system

### DIFF
--- a/recipe/src/module.rs
+++ b/recipe/src/module.rs
@@ -101,6 +101,13 @@ impl<'a> Module<'a> {
     pub fn generate_akmods_info(&'a self, os_version: &str) -> AkmodsInfo {
         trace!("generate_akmods_base({self:#?}, {os_version})");
 
+        // Akmods don't support the latest tag.
+        let os_version = if os_version == "latest" {
+            "39"
+        } else {
+            os_version
+        };
+
         let base = self
             .config
             .get("base")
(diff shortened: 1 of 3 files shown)


xynydev (approved):

fiftydinar:
The issue with this is that it's not dynamic, you would need to change this every time new Fedora release comes in.

bayou-brogrammer:
> The issue with this is that it's not dynamic, you would need to change this every time new Fedora release comes in.

it is not meant to be dynamic. I think its ok to have hardcoded values. We can just come back and revisit this on the next release. Releases happen on a ~6month cadence.

bayou-brogrammer:
it is not meant to be dynamic. I think its ok to have hardcoded values. We
can just come back and revisit this on the next release. Releases happen on
a ~6month cadence.

gmpinder:
We get the OS version from labels. Is there a reason why that's not working for you? I'm with @fiftydinar on this, we shouldn't hard code this in and instead find a more reliable way to get the information we need

xynydev (changes requested):
Ok, yeah, I didn't look at it so closely. I agree.

fiftydinar:
@bayou-brogrammer it would be maybe a good idea to land last "default to main" fix in a separate PR.

bayou-brogrammer:
> @bayou-brogrammer it would be maybe a good idea to land last "default to main" fix in a separate PR.

crap didnt realize i was still on this branch

PR #110, opened 2024-03-01
fix: use container skopeo

bayou-brogrammer:
the `os_version` is defaulting to the `image_tag` inside containers and causing our template to use latest tag


src/commands/template.rs
```diff
@@ -26,6 +26,10 @@ pub struct TemplateCommand {
     #[clap(skip)]
     #[builder(default, setter(into, strip_option))]
     build_id: Option<Uuid>,
+
+    #[clap(skip)]
+    #[builder(default)]
+    os_version: String,
```
gmpinder, line 32: This won't work when you use the `template` command. This will end up being an empty string and cause the template to be useless if a user is wanting to use the akmods module in their recipe.

template/templates/Containerfile.j2
```diff
@@ -1,4 +1,4 @@
-{%- let os_version = recipe.get_os_version() %}
+{%- let os_version = self.os_version %}
```
gmpinder, line 1: I don't think this is valid syntax. You would need to add the `os_version` to add this to `ContainerFileTemplate` and then you can remove this line entirely.

gmpinder (changes requested):

template/templates/Containerfile.j2
gmpinder: We can just delete this. I'm an idiot, you're already doing that.
```suggestion
```

src/commands/build.rs
```diff
@@ -451,6 +464,37 @@ impl BuildCommand {
...
     }
+
+    fn get_os_version(
+        &self,
+        build_strat: Rc<dyn BuildStrategy>,
+        recipe: &Recipe,
+    ) -> Result<String> {
+        trace!("BuildCommand::get_os_version({recipe:#?})");
+
+        let scopeo_output = build_strat.inspect(&recipe.base_image, &recipe.image_version)?;
```
xynydev, line 475: It's called `skopeo` with a `k`. https://github.com/containers/skopeo

xynydev: Also, I can't seem to find a definition for the `inspect` method used here(?)

src/strategies.rs
```diff
@@ -38,12 +104,86 @@ pub trait BuildStrategy {
...
+
+    let os_version = match entry {
+        None => {
+            let inspection =
+                INSPECT_STRATEGY.get_labels(&recipe.base_image, &recipe.image_version)?;
+
+            let os_version = inspection.get_version().unwrap_or_else(|| {
+                warn!("Version label does not exist on image, using version in recipe");
+                recipe.image_version.to_string()
+            });
```
gmpinder, line 149: Now that we have a way to dynamically choose an inspection strategy that supports not having skopeo, I'm thinking we should just error if the label for the os version doesn't exist on the image.

src/strategies.rs
```diff
@@ -38,12 +104,86 @@ pub trait BuildStrategy {
...
+            blue_build_utils::check_command_exists("skopeo"),
+            blue_build_utils::check_command_exists("docker"),
+            blue_build_utils::check_command_exists("podman"),
+            blue_build_utils::check_command_exists("buildah"),
+        ) {
+            (Ok(_skopeo), _, _, _) => Arc::new(SkopeoStrategy),
+            (_, Ok(_docker), _, _) => Arc::new(DockerStrategy),
+            (_, _, Ok(_podman), _) => Arc::new(PodmanStrategy),
+            (_, _, _, Ok(_buildah)) => Arc::new(BuildahStrategy),
+            _ => bail!("Could not determine inspection strategy. You need eiterh skopeo, docker, podman, or buildah"),
```
gmpinder, line 180:
```suggestion
            _ => bail!("Could not determine inspection strategy. You need either skopeo, docker, podman, or buildah"),
```

gmpinder (dismissed):

gmpinder (approved):

PR #114, opened 2024-03-02
fix: Remove check for specific branches for signing

gmpinder:


gmpinder:
Diff might be hard to read, but I moved the check for a cosign priv/pub key to the top of the match statement so that it is the first thing that is tried everywhere. I also removed all checks for which branch it is building on so that we can sign for any branch build.

PR #115, opened 2024-03-06
fix: update path in comments and README

davemccrea:


gmpinder:
Didn't notice, thanks for catching that!

gmpinder (approved):

PR #116, opened 2024-03-06
refactor!: Rename `template` to `generate` and move `rebase/upgrade` under `switch`

### LINKED ISSUE #159: fix: check if current transaction is ongoing in rebase command
When automatic upgrades are enabled users might use the rebase command while a transaction is ongoing. Currently in main the `rpm-ostree rebase` subprocess terminates with `Transaction in progress:`. This is bad UX since this leads users to have to restart the rebuild process which can take minutes.

I think the command should either check the `rpm-ostree` status initially and inform the user or much simpler, when the final rebase subprocess terminates inform the user about the built file in `/etc/bluebuild`. 

Tell me what you think about this i could work on this if you think this is an issue as well.

gmpinder:
This updates the `template` subcommand to be `generate`. The `template` usage will continue to work as an alias to `generate`. A new `switch` command is added that will manage both `rpm-ostree rebase` and `rpm-ostree upgrade` and is fully replacing the respective subcommands as a breaking change.

The new `switch` command is under the feature flag `switch` and will currently only build for the `main` branch builds until it is moved as a default feature (`v0.9.0`).

Closes #159


gmpinder:
Still need to add logic to auto-determine `rpm-ostree` command and update the README to match these changes.

fiftydinar (approved):
Looks good from my limited understanding

PR #123, opened 2024-03-17
fix: Remove tmpfs for /var and /tmp

gmpinder:


gmpinder:
So I'm running into an issue where something isn't being parsed out properly. I'm not sure if it's because of the escaping or what. I'm thinking maybe we have like a separate image that is created that holds all of our export "util" scripts. This can be managed and updated on its own and we would mount it and source the scripts. This would allow the use of comments and more complex syntax that wouldn't be allowed by just printing the script directly in the Containerfile.

gmpinder:
Decided to not make it to complicated for now and save the script injection for later (see https://github.com/blue-build/cli/issues/124).

PR #126, opened 2024-03-17
fix: Allow docker driver to properly use cache

gmpinder:
This fix involves creating a new function for the `BuildDriver` trait called `build_tag_push`. In order to get the proper logic in place to make use of `docker buildx`, I had to create a separate function that would construct the build command to include all of the tags necessary for pushing. A default implementation of `build_tag_push` will be used for `podman` and `buildah` which was originally from the build command's functions. Now that we have custom logic for docker builds, we can take advantage of using the GitHub cache features without having the `--load` arg which had a big negative effect on build times. We can now also use docker for creating local `oci-archive` tarballs for local rebasing. Making use of the `oci-archive` will require the user to create a `docker-container` builder as it is not supported on the standard `docker` builder.

https://docs.docker.com/build/exporters/oci-docker/

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -219,7 +219,6 @@ impl BlueBuildCommand for BuildCommand {
 
         if self.push {
             blue_build_utils::check_command_exists("cosign")?;
-            blue_build_utils::check_command_exists("skopeo")?;
             check_cosign_files()?;
         }
 
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -312,61 +312,53 @@ impl BuildCommand {
         trace!("BuildCommand::generate_full_image_name({recipe:#?})");
         info!("Generating full image name");
 
-        let image_name = if let Some(archive_dir) = &self.archive {
-            format!(
-                "{}/{}.{ARCHIVE_SUFFIX}",
-                archive_dir.to_string_lossy().trim_end_matches('/'),
-                recipe.name.to_lowercase().replace('/', "_"),
-            )
-        } else {
-            match (
-                env::var(CI_REGISTRY).ok().map(|s| s.to_lowercase()),
-                env::var(CI_PROJECT_NAMESPACE)
-                    .ok()
-                    .map(|s| s.to_lowercase()),
-                env::var(CI_PROJECT_NAME).ok().map(|s| s.to_lowercase()),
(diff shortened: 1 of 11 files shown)


src/commands/build.rs
```diff
@@ -286,121 +312,60 @@ impl BuildCommand {
...
+            }
+            (None, None, None, Some(github_repository_owner), None, None) => {
+                trace!("GITHUB_REPOSITORY_OWNER={github_repository_owner}");
+                warn!("Generating Github Registry image");
+                format!("ghcr.io/{github_repository_owner}/{}", &recipe.name)
+            }
+            _ => {
+                trace!("Nothing to indicate an image name with a registry");
+                if self.push {
+                    bail!("Need '--registry' and '--registry-path' in order to push image");
```
gmpinder, line 359:
```suggestion
                    bail!("Need '--registry' and '--registry-namespace' in order to push image");
```

src/drivers.rs
```diff
@@ -240,7 +302,7 @@ impl Driver<'_> {
     }
 
     fn determine_inspect_driver() -> Result<Arc<dyn InspectDriver>> {
-        trace!("Strategy::determine_inspect_strategy()");
+        trace!("Driver::determine_inspect_strategy()");
```
gmpinder, line 305:
```suggestion
        trace!("Driver::determine_inspect_driver()");
```

src/drivers.rs
```diff
@@ -257,7 +319,7 @@ impl Driver<'_> {
     }
 
     fn determine_build_driver() -> Result<Arc<dyn BuildDriver>> {
-        trace!("Strategy::determine_build_strategy()");
+        trace!("Driver::determine_build_strategy()");
```
gmpinder, line 322:
```suggestion
        trace!("Driver::determine_build_driver()");
```

src/drivers/opts/build.rs
```diff
@@ -0,0 +1,24 @@
+use std::borrow::Cow;
+
+use typed_builder::TypedBuilder;
+
+#[derive(Debug, Clone, TypedBuilder)]
+pub struct BuildTagPushOpts<'a> {
```
gmpinder, line 6: Should add doc comments

gmpinder:
Ok I think this is ready to be looked at

gerblesh (approved):
lgtm

PR #132, opened 2024-03-23
fix: Copy bins and keys with mounts for ostree commit

gmpinder:
I've been investigating more into how ostree works and how it relates to running `ostree container commit` for each layer. I've decided to move our pre-installed bins and public keys into their own stages and then bind mount them into a `RUN` instruction so that we can just use `cp` to get the files into the image and then call `ostree container commit`. Now all of our layers in the image (after the base image) will be in the ostree commit tree.


gmpinder:
Integration tests will fail until https://github.com/blue-build/modules/pull/174 is merged in.

fiftydinar:
> Integration tests will fail until [blue-build/modules#174](https://github.com/blue-build/modules/pull/174) is merged in.

Merged

xynydev (approved):

PR #139, opened 2024-03-26
feat: Improve logging output

gmpinder:


gmpinder:
I decided to increase the amount of information you get the more verbose you set the logs to be. Here are the formats:

- Error, Warn, and Info level
image

- Debug Level
image

- Trace Level
image

What's nice is, now the trace logging will display the line number in the code which will perfect for debugging!

gmpinder:
Thinking about it, probably don't want to have white for the time in case users use a white background

gmpinder:
Here's standard fg color for time 

image

gmpinder:
And now with some proper padding 

image

xynydev (approved):

xynydev (commented):
Actually, uhh, why is most of the text orange in your screenshot? Shouldn't color be used as a highlight and not just added to everything for the sake of it. Or is this a setting of your terminal?

fiftydinar:
> Actually, uhh, why is most of the text orange in your screenshot? Shouldn't color be used as a highlight and not just added to everything for the sake of it. Or is this a setting of your terminal?

Orange color is usually used to indicate warning or attention, along with yellow for caution.

I think that terminals automatically switch colors from black to white & vice-versa when switching the theme.
So I'm not sure if there is a need to make everything orange to comply with most themes.

gmpinder:
> Actually, uhh, why is most of the text orange in your screenshot? Shouldn't color be used as a highlight and not just added to everything for the sake of it. Or is this a setting of your terminal?

The orange is my personal color scheme. The parts of the log that aren't orange are the colors that are added.

PR #158, opened 2024-04-12
chore: Cleanup images and use hash for exports tag

gmpinder:
I noticed that making changes to the exports script before a release could cause modules to not build properly if breaking changes got pushed out. To prevent this, I'm making it so that the hash of the commit is put in the tag for the exports script image and that the CLI tool will use that hash when building the `Containerfile`.


gmpinder:
Ok this is cleaned up now. Lots of trial and error in this PR

xynydev (approved):

PR #182, opened 2024-05-13
feat(experimental): Build multiple recipes in parallel

gmpinder:
The `build` subcommand can now take in any number of recipe files and will build them all in parallel. Along with this new ability, I've added a way to easily distinguish which part of the build log belongs to which recipe. Check out the `docker_build` action of this PR for an example.

gif

## Tasks

- [x] Make build log follow same pattern as normal logs to keep things consistent
- [x] Update color ranges based on @xynydev 's feedback
- [x] Deal with ANSI control characters in log output
- [x] Add [`indicatif`](https://crates.io/crates/indicatif) to make logs look nicer
- [x] Add ability to print logs to a file


Cargo.toml
```diff
@@ -81,6 +81,7 @@ serde_json.workspace = true
 serde_yaml.workspace = true
 typed-builder.workspace = true
 uuid.workspace = true
+rayon = "1.10.0"
```
gmpinder, line 84: Move this

src/commands/build.rs
```diff
@@ -120,133 +121,95 @@ impl BlueBuildCommand for BuildCommand {
...
+                vec![(CONTAINER_FILE.to_string(), recipe_path.join(RECIPE_FILE))]
             } else {
                 warn!("Use of {CONFIG_PATH} for recipes is deprecated, please move your recipe files into {RECIPE_PATH}");
-                legacy_path.join(RECIPE_FILE)
+                vec![(CONTAINER_FILE.to_string(), legacy_path.join(RECIPE_FILE))]
             }
+        },
+        |recipes| {
+            recipes.into_par_iter().filter_map(|recipe| {
+                Some((format!("{CONTAINER_FILE}.{}", recipe.file_stem()?.to_str()?), recipe))
```
gmpinder, line 147: Maybe use the recipe path instead and replace `/` with `_`

src/drivers/docker_driver.rs
```diff
@@ -215,6 +221,8 @@ impl BuildDriver for DockerDriver {
                     }
                 }
 
+                final_image.push_str(image);
```
gmpinder, line 224: Use the first tagged image instead

utils/src/command_output.rs
```diff
@@ -1,12 +1,63 @@
...
+}
+
+impl CommandExt for Command {
+    fn status_log_prefix<T: AsRef<str>>(&mut self, log_prefix: &T) -> Result<ExitStatus> {
+        let mut child = self.stdout(Stdio::piped()).stderr(Stdio::piped()).spawn()?;
+
+        let mut rng = rand::thread_rng();
+        let red: u8 = rng.gen_range(80..=240);
+        let green: u8 = rng.gen_range(80..=240);
+        let blue: u8 = rng.gen_range(80..=240);
```
gmpinder, line 30: @xynydev I could use your artistic eye in getting a good range of colors here. I just picked a range that wouldn't create something too light that you couldn't distinguish between the builds and something too dark for dark terminal users (such as myself).

src/commands/build.rs
gmpinder: Went with a hash of the recipe file path instead so each Containerfile will have only one recipe.

utils/src/command_output.rs
xynydev: Could you maybe tell me where this is used and/or send a screenie?

xynydev: Also, for random color generation, setting a standard saturation and lightness, while randomizing the hue, is what I think could work best. Generative art, eh?

gmpinder: Yeah this will be used for command output for builds. It will prepend each line with the image that is being built so that users can follow along easier. Let me get some examples

gmpinder: image

gmpinder: Here's one from a different terminal 
image

gmpinder: > Also, for random color generation, setting a standard saturation and lightness, while randomizing the hue, is what I think could work best. Generative art, eh?

I'll look into this. The library I use does RGB, but I know I could do some math for HSL.

gmpinder:
Ok so trying this out, it would seem that although GHAs support using truecolor, Gitlab CI doesn't and makes it look weird. So I'm going to try using standard ANSI colors to keep things as compatible as possible.

gerblesh (approved):
lgtm so far

xynydev:
> Ok so trying this out, it would seem that although GHAs support using truecolor, Gitlab CI doesn't and makes it look weird. So I'm going to try using standard ANSI colors to keep things as compatible as possible.

This is probably something that shouldn't be used in a CI with build matrixing capabilities, but could be useful for some in local builds or custom CI systems I guess. 

The logs being interwoven like that could get cluttered, though the custom colors help, maybe adding an option to write the logs to some directory instead would make this even better?

gmpinder:
> This is probably something that shouldn't be used in a CI with build matrixing capabilities, but could be useful for some in local builds or custom CI systems I guess.

My next goal after this is to make use of `buildx bake`. There are situations where some stages can be shared amongst builds like compiling stages and bake would help to ensure that they do. Currently this might or might not happen when spawning docker processes during my own testing. I did try it out locally so I guess I could try adding it to this PR now that I know how that works

gmpinder:
> The logs being interwoven like that could get cluttered, though the custom colors help, maybe adding an option to write the logs to some directory instead would make this even better?

Yeah that was the reason for the colors. Earthly has a similar pattern for their matrix builds. I think it would be reasonable/fairly easy to add log redirection too. Helps that I already created a system to handle logs

gmpinder:
I'll hold off on the use of buildx bake for this particular PR and add that in later. I will still work on getting logs out

PR #187, opened 2024-05-29
fix: Remove hard requirement for login creds to be able to push

gmpinder:
Related to https://github.com/blue-build/github-action/issues/48


gerblesh (commented):
LGTM

gerblesh:
looks like CI is having some hiccups but the PR looks solid

gmpinder:
> looks like CI is having some hiccups but the PR looks solid

Yeah that's been happening recently and I'm not sure why. Might be time to rotate the earthly satellite.

PR #192, opened 2024-06-09
feat(iso): Create generate-iso command

gmpinder:
## Tasks

- [x] Add ctrl-c handler to kill spawned children
- [x] add more args to support all variables
- [x] Add integration test


fiftydinar:
Can't see from the commits, does this uses this action?

https://github.com/JasonN3/build-container-installer

Bundling flatpaks inside the ISO from the `default-flatpaks` module recipe would be also great.

However, I find that it doesn't work properly in my testing for some reason (flatpaks installed, but they miss refs, hence they failed to update).
However, I use custom named `flathub-system` remote instead of `flathub` so maybe that's what contributed to it.

gmpinder:
> Can't see from the commits, does this uses this action?
> 
> https://github.com/JasonN3/build-container-installer
> 
> Bundling flatpaks inside the ISO from the `default-flatpaks` module recipe would be also great.
> 
> However, I find that it doesn't work properly in my testing for some reason (flatpaks installed, but they miss refs, hence they failed to update).
> However, I use custom named `flathub-system` remote instead of `flathub` so maybe that's what contributed to it.

I'm still working on the implementation, but yes I'll be using the image from that action to build the ISO. I've got a lot of testing to do and I'd appreciate y'all testing it out too when I get it into a decent state

src/commands/generate_iso.rs
```diff
@@ -0,0 +1,98 @@
...
+                .build(),
+            RunOptsEnv::builder()
+                .key("VARIANT")
+                .value(&self.variant)
+                .build(),
+            RunOptsEnv::builder()
+                .key("DNF_CACHE")
+                .value("/cache/dnf")
+                .build(),
+        ];
```
gmpinder, line 83: Based on testing, it seems setting these as env vars doesn't work and will need to be passed in as args like in the action

fiftydinar:
It seems that latest Fedora 40 images require `bootupd` to be installed & `bootupctl backend generate-update-metadata` command to be done inside the image. The command updates the bootloader metadata which would be further used by `bootc` in installer to update the bootloader data inside the system to latest version. That's how I understand it.

Maybe old Anaconda versions prior to latest 1.2.1 install the older bootloader version, which is not compatible with the kernel. 1.2.1 adds the support for using `bootc` to circumvent this issue. This is just a speculation, I'm not sure.

I made the issue in upstream ISO generator repo:
https://github.com/JasonN3/build-container-installer/issues/127

PR #194, opened 2024-06-22
fix: Support other signals properly

gmpinder:


gmpinder:
Ok so the issue is because of the GitlabRunners (https://gitlab.com/gitlab-org/gitlab-runner/-/issues/37780). The changes I made in this PR though is still something I think we're going to keep as it gives us finer grain control of signals for future features.

PR #197, opened 2024-07-05
refactor: Create SigningDriver and CiDriver

gmpinder:
This adds the driver for handling signing. It will make use of cosign if available and fall back to running the cosign container in docker or podman.


fiftydinar:
Is this going to pair well with bash `signing` module, or that module can be even ditched for this one in cli?

gmpinder:
> Is this going to pair well with bash `signing` module, or that module can be even ditched for this one in cli?

This is mostly just for performing all the various cosign operations during build/init to make things easier if a user doesn't have cosign installed. It's more of a refactor than anything

PR #206, opened 2024-07-25
feat: support VanillaOS images

xynydev:
I have successfully used a CLI installed with these modifications to build an image and publish it on [`ghcr.io/xynydev/vanillin`](https://github.com/xynydev/vanillin). I have switched to using the image in my VanillaOS VM and am currently faced with this error:
image

However, it was possible to ignore the error and try to log in, which froze the system, though the systemd unit for `default-flatpaks` ran...

Unsolved:
- How to dynamically switch between `ostree` and `vanilla` templates
- How to deal with Vanilla not providing `org.opencontainers.image.version` and not even really having versions in the same way as Fedora
   - As an aside, the version variable probably shouldn't be an int, since it might be the case that not all base images have simple numerical versions
- Where to set the correct image name in `abroot.json`, in the `signing` module or the CLI template
   - Further discussion: does it make sense to have a `signing` module that everyone always uses, or should we "upstream" that into CLI
   - For now, I solved this with a quick [script](https://github.com/xynydev/vanillin/blob/main/files/scripts/abroot.sh)

Links:
- https://github.com/Vanilla-OS/custom-image/blob/main/recipe.yml
- https://github.com/Vanilla-OS/vib-fsguard
- https://github.com/Vanilla-OS/desktop-image


fiftydinar:
VanillaOS v1 users have to reinstall to get into VanillaOS v2, since it got refactored, as far as I'm aware.

Is there any guarantee that this scenario won't happen when migrating from v2 to v3?

If it does happen, how would we handle those scenarios?

xynydev:
> VanillaOS v1 users have to reinstall to get into VanillaOS v2, since it got refactored, as far as I'm aware.

Yup, I think so too. v1 is built on Ubuntu and ABRoot v1, while v2 (Orchid) is on Debian and ABRoot v2.

> Is there any guarantee that this scenario won't happen again when migrating from v2 to v3?

No. Actually, I think that there might be a guarantee that this scenario _will_ happen when a hypothetical v3 comes, because the thing about major version upgrades is that they make breaking changes. However, as long as Vanilla stays on a Debian / ABRoot base that is compatible with Orchid, there should be no problems. 

> If it does happen, how would we handle those scenarios?

Depends on how VanillaOS handles them. But if the build processes are incompatible, I guess we could support both, at least for a while. There's not much else that would be our business to do.

template/src/lib.rs
```diff
@@ -13,7 +13,11 @@ use uuid::Uuid;
 pub use askama::Template;
 
 #[derive(Debug, Clone, Template, TypedBuilder)]
-#[template(path = "Containerfile.j2", escape = "none", whitespace = "minimize")]
+#[template(
+    path = "Containerfile.vanilla.j2",
+    escape = "none",
+    whitespace = "minimize"
+)]
 pub struct ContainerFileTemplate<'a> {
```
gmpinder, line 21: Rather than changing the existing template struct, I would suggest making a brand new template struct.

src/drivers.rs
```diff
@@ -389,12 +389,7 @@ impl Driver<'_> {
                     .build();
                 let inspection = INSPECT_DRIVER.get_metadata(&inspect_opts)?;
 
-                let os_version = inspection.get_version().ok_or_else(|| {
-                    miette!(
-                        help = format!("Please check with the image author about using '{IMAGE_VERSION_LABEL}' to report the os version."),
-                        "Unable to get the OS version from the labels"
-                    )
-                })?;
+                let os_version = inspection.get_version().unwrap_or(0);
```
gmpinder, line 392: Let's leave this function alone and instead call a different function for whatever the vanilla template needs. We want to try to keep the fedora templates and the vanilla templates separate as possible.

template/templates/modules/modules.j2
```diff
@@ -22,22 +23,18 @@ RUN \
...
-  && ostree container commit
+  /tmp/scripts/run_module.sh '{{ module.module_type }}' '{{ module.print_module_context() }}'
       {%- endif %}
     {%- endif %}
   {%- endfor %}
 {% endmacro %}
-{% macro stage_modules_run(modules_ext, os_version) %}
+
+
+{% macro ostree_modules_run(modules_ext, os_version) %}
```
gmpinder, line 34: The original module macros here should remain unchanged. I would suggest making a new module macro for the vanilla OS.

gmpinder (changes requested):

template/src/lib.rs
xynydev: I agree. I just did it like this for testing purposes, and because I kind of wish(ed) that there would be some way to dynamically change the template file without duplicating the struct.

src/drivers.rs
xynydev: I agree. I just did it like this because I couldn't figure out any easier way to fix this error. We should have some way to detect and declare what the base image is, and call different functions based on that, and remove the hard requirement for `os_version` in the tagging system, etc.

template/templates/modules/modules.j2
xynydev: I split this to generic / ostree macros, because only rpm-ostree module calls require the things added by the ostree macro, VanillaOS, stages, etc., do not and probably never will require anything like `ostree container commit`. If there is some sort of integration needed like the we have with the `akmods` module, it would be trivial to split `generic` into `generic` and `vanilla` and switch the vanilla template to use the new macro.

I'm envisioning that we'd also ship a `generic` base image type, which would not add any OS-specific things to the Containerfile, and could thus be usable for basically any operating system that supports OCI images as a distribution mechanism (without extra work from us, but with extra work from the custom image maintainer).

mirkobrombin:
> VanillaOS v1 users have to reinstall to get into VanillaOS v2, since it got refactored, as far as I'm aware.
> 
> Is there any guarantee that this scenario won't happen again when migrating from v2 to v3?
> 
> If it does happen, how would we handle those scenarios?

That's not gonna happening.

xynydev:
I just removed all the custom changes from my BlueBuild-generated custom image, built, pushed, and updated, and still got hit with the FsGuard issue. Just to make sure, I also built the Vib custom image template locally, pushed to the same image with a different tag, and rebased to that, and as expected there was no FsGuard pop-up. 

The BlueBuild-generated Containerfile has basically the exact same lines for FsGuard as the Vib-generated one. The issue is probably with some of the "boilerplate" in the image, so then next step I guess is to slowly remove that until it starts working...

xynydev:
Those previous two `chore` commits did not help this issue at all. Maybe a [GitHub attestation](https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds) is required for an image to go through FsGuard properly? I can't find the part in the [`vib-build.yml`](https://github.com/Vanilla-OS/custom-image/blob/main/.github/workflows/vib-build.yml) where the attestation is done, but I doubt it would work with local test builds anyway, so I'd have to move to CI builds for testing.

mirkobrombin:
> I just removed all the custom changes from my BlueBuild-generated custom image, built, pushed, and updated, and still got hit with the FsGuard issue. Just to make sure, I also built the Vib custom image template locally, pushed to the same image with a different tag, and rebased to that, and as expected there was no FsGuard pop-up.
> 
> The BlueBuild-generated Containerfile has basically the exact same lines for FsGuard as the Vib-generated one. The issue is probably with some of the "boilerplate" in the image, so then next step I guess is to slowly remove that until it starts working...

FsGuard only checks for the binaries listed in its hash table. If any of the file hashes changes, FsGuard will complain. FsGuard generation should always happen as the final step of the build process.

xynydev:
Ok. Last time I was in a rush, so I made an oopsie. I used the `arm64` URL for FsGuard instead of the `amd64` URL. I was correct last time, but the verification still failed thanks to that detail. 

ie., it actually works now! :tada:

template/src/lib.rs
xynydev: Solved here: [622f895](https://github.com/blue-build/cli/pull/206/commits/622f89599ffafc4c9d459539f6b257737ee7a7ed)

xynydev:
First successful CI build: https://github.com/xynydev/vanillin/actions/runs/10282268597

`base-image-type: vanilla` is required to specify the space image type currently.

xynydev:
Other issues:
- When not building with `squash: true`, upgrades are broken for images built by BlueBuild
   - `abroot` just says "No upgrade available."
   - Running `abroot uprgade --verbose` gives this as the error: `MANIFEST_UNKNOWN message:OCI index found, but Accept header does not support OCI indexes`

gmpinder:
Closing as the code is very out of date and hasn't been worked on in over a year.

xynydev:
Good. I mean, nobody really cares about this, but _it would be so easy to do_ since it's also just OCI, feels kind of like a wasted opportunity. But then again, where's the waste if no-one wants it... 

This PR def aint a good starting point for implementing it anyways.

PR #209, opened 2024-07-27
fix: add typespec schemas for cli modules, remove modules.json (not needed anymore)

xynydev:
The website build process now uses the GitHub API to generate a global `modules.json`. If you decide to move the directory containing all the modules, please tell me, or make a PR changing [this line](https://github.com/blue-build/website/blob/9eb198c4e532e82e70b980da955b87f39ce663b4/astro.config.mjs#L102). If you don't, I'll find out anyways, because the website builds will break. 

Also, whenever updating the modules, make sure the schema is updated too to match the current state of the module. If you need help with writing [TypeSpec](https://typespec.io/), consult me, but you probably wont, since it's just a type system kind of like TypeScript's or Rust's .


gmpinder (approved):
Seems easy enough. Thanks for getting this taken care of!

PR #214, opened 2024-08-08
chore: Move files from `/usr/etc/` to `/etc/` in build-time

fiftydinar:
Take a look at this issue for more details:

https://github.com/blue-build/modules/issues/314

Idk if anything else needs to be done, please tell

Should be tested for regressions too


gmpinder (approved):
Is this really it? Easy approval

fiftydinar:
Don't see anything else that needs to be done, so easy-peasy I guess

PR #229, opened 2024-09-19
chore: Migrate script module name to scripts

fiftydinar:
Only fixed the comment in code,

idk if something else should be done in order to support both `script` & `scripts` module name in stages

Parent PR:
https://github.com/blue-build/modules/pull/334


gmpinder:
There's some recipes in the `integration-tests` dir that could be updated too.

PR #230, opened 2024-09-21
refactor: Swtich to using bon for builder pattern

gmpinder:


process/drivers/opts/ci.rs
```diff
@@ -1,24 +1,24 @@
...
 
-#[derive(Debug, Clone, TypedBuilder)]
+#[derive(Debug, Clone, Builder)]
 pub struct GenerateTagsOpts<'scope> {
     pub oci_ref: &'scope Reference,
 
-    #[builder(default, setter(into))]
-    pub alt_tags: Option<Vec<Cow<'scope, str>>>,
+    #[builder(into)]
+    pub alt_tags: Option<Vec<String>>,
```
gmpinder, line 11: Still unsure what the best way is to handle vecs and reduce memory footprint

gmpinder:
I'm unable to convert traits over to using `bon` to create a builder pattern for traits specifically. Looks like it's something to look forward to https://github.com/elastio/bon/issues/123

PR #233, opened 2024-09-29
feat: Add platform arg to force building a specific architecture

gmpinder:


process/drivers/opts/inspect.rs
```diff
@@ -2,11 +2,14 @@ use std::borrow::Cow;
...
+use crate::drivers::types::Platform;
+
 #[derive(Debug, Clone, Builder)]
 pub struct GetMetadataOpts<'scope> {
     #[builder(into)]
     pub image: Cow<'scope, str>,
 
     #[builder(into)]
     pub tag: Option<Cow<'scope, str>>,
+    pub platform: Platform,
```
gmpinder, line 14: Add builder default

process/drivers/traits.rs
```diff
@@ -215,9 +216,9 @@ pub trait SigningDriver {
         let inspect_opts = GetMetadataOpts::builder().image(image_name);
 
         let inspect_opts = if let Some(ref tag) = opts.tag {
-            inspect_opts.tag(tag.as_ref() as &str).build()
+            inspect_opts.tag(&**tag).platform(opts.platform).build()
         } else {
-            inspect_opts.build()
+            inspect_opts.platform(opts.platform).build()
```
gmpinder, line 221: Move platform calls up to 216

process/drivers/docker_driver.rs
```diff
@@ -130,6 +132,10 @@ impl BuildDriver for DockerDriver {
         let status = cmd!(
             "docker",
             "build",
+            if matches!(opts.platform, Platform::LinuxAmd64 | Platform::LinuxArm64) => [
```
gmpinder, line 135: Let's do not matches Native instead in case we add more platforms

PR #237, opened 2024-10-06
fix: switch cosign registry from GCR to GHCR

bobcallaway:
This changes the Dockerfile to pull the cosign container image from GHCR instead of Google Cloud. This helps the Sigstore team manage their cloud spend (as GHCR is provided for free and Google Cloud Artifact Registry is not).

Note the container hash does not change and images are posted to both locations upon cosign's release process.


gmpinder (approved):
Huh, this was never intended to use Google Cloud. I think I must have typo'd this and it was just a happy mistake that it worked.

bobcallaway:
This was the normal instructions on sigstore.dev - so probably not a mistake on your part. thanks for merging!

PR #239, opened 2024-10-12
feat: Add validation command

gmpinder:
Adds JSON Schema validation for the recipes. This will print out a list of errors encountered when the recipe does not match the schema.

In order to get the schema to validate properly, we had to include the requirement of having `source` set to `local` on a locally built module. The `validate` command will be put under a feature flag until the `v0.9` release to prevent breaking user's configs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 .vscode/
 result*
 .direnv/
+.arg
+.secret
 
 cosign.key
 !test-files/keys/cosign.key
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,6 +58,7 @@ dependencies = [
  "cfg-if",
  "getrandom",
  "once_cell",
+ "serde",
  "version_check",
  "zerocopy",
 ]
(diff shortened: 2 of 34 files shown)


xynydev:
Should we implement recipe v2 next, and the announce all the breaking changes at once? 
Like:
- You should now refactor your recipes (which will automatically opt in to recipe validation)
- But if you don't want to do that, you don't need to (yet?)
   - But if you want to opt in to recipe validation for recipe v1, you still need to add `source: local` for local modules

If we want a `version:` key for example, would it be possible to make the presence of `version: 1` opt in to recipe validation for v1 recipes? Then to use v2 one would need to specify `version: 2`, also enabling validation.

gmpinder:
> If we want a `version:` key for example, would it be possible to make the presence of `version: 1` opt in to recipe validation for v1 recipes? Then to use v2 one would need to specify `version: 2`, also enabling validation.

I think this is a great idea to opt-in! The user shouldn't need to know the schema url to take advantage of validation (though it would be needed if they want the validation in their editors).

gmpinder:
> Should we implement recipe v2 next, and the announce all the breaking changes at once?

Yeah I think that would be a good step. I think getting `validate` and `new`/`init` commands should be the last things before the `v0.9` release. I'd also like to fix the whole distrobox issue #160 too, but could use some help with that since I don't use distrobox much.

PR #264, opened 2024-11-14
fix: Include base image information in labels

gmpinder:


template/templates/Containerfile.j2
```diff
@@ -2,7 +2,7 @@
 {%- include "stages.j2" %}
 
 # Main image
-FROM {{ recipe.base_image }}:{{ recipe.image_version }} AS {{ recipe.name|replace('/', "-") }}
+FROM {{ recipe.base_image }}@{{ base_digest }} AS {{ recipe.name|replace('/', "-") }}
```
yacoob, line 5: Thank you for including this suggestion! :smile_cat:

PR #280, opened 2024-12-01
chore: Assure that `get_json_array` outputs compact `json` output

fiftydinar:
This fixes the issue with `files` module in this [PR](https://github.com/blue-build/modules/pull/369), where brackets are included in an array.


gmpinder (approved):
Shit, I thought I'd put that in there

PR #288, opened 2024-12-07
chore: Copy signing keys to `/etc/` only

fiftydinar:
Supplements main PR: https://github.com/blue-build/modules/pull/375

Tests & it works, can be merged.

### DIFF
--- a/integration-tests/test-repo/recipes/recipe.yml
+++ b/integration-tests/test-repo/recipes/recipe.yml
@@ -30,7 +30,8 @@ modules:
       - firefox-langpacks
 
   - type: signing
-
+    source: ghcr.io/blue-build/modules:signing-switch-fully-to-etc
+    
   - type: test-module
     source: local
 
--- a/integration-tests/test-repo/recipes/recipe.yml
+++ b/integration-tests/test-repo/recipes/recipe.yml
@@ -30,7 +30,6 @@ modules:
       - firefox-langpacks
 
   - type: signing
-    source: ghcr.io/blue-build/modules:signing-switch-fully-to-etc
     
   - type: test-module
     source: local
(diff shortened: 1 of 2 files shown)


gmpinder (approved):
The changes look good, but it seems like the signing module is causing failures. I think that will have to be updated before this can be merged in.

fiftydinar:
> The changes look good, but it seems like the signing module is causing failures. I think that will have to be updated before this can be merged in.

Hmm, error implies that keys are not copied to `/etc/pki/containers/`.

Maybe that's only in Ublue image. I would like to know if that issue is gone in vanilla Fedora based images.

fiftydinar:
It would be good to use the `signing` module from my PR temporarily to test this, then revert it back.
Is that good?

fiftydinar:
> It would be good to use the `signing` module from my PR temporarily to test this, then revert it back. Is that good?

I confirm it works now.

xynydev (approved):

gmpinder (approved):

gmpinder:
> > It would be good to use the `signing` module from my PR temporarily to test this, then revert it back. Is that good?
> 
> I confirm it works now, so it can be merged.

I'm not comfortable merging this in until https://github.com/blue-build/modules/pull/375 is merged. I run my builds of the `main` branch and this will fail them without the module changes.

fiftydinar:
> > > It would be good to use the `signing` module from my PR temporarily to test this, then revert it back. Is that good?
> > 
> > 
> > I confirm it works now, so it can be merged.
> 
> I'm not comfortable merging this in until [blue-build/modules#375](https://github.com/blue-build/modules/pull/375) is merged. I run my builds of the `main` branch and this will fail them without the module changes.

Merged

PR #296, opened 2024-12-12
chore: Use consistent syntax for getting information from os-release

fiftydinar:
This will make it work for Ubuntu versions, where it's `24.04` or similar. Or if some distros use letters for VERSION_IDs.

Please tell me if I escaped quotes correctly in `drivers.rs`

### DIFF
--- a/process/drivers.rs
+++ b/process/drivers.rs
@@ -282,7 +282,7 @@ fn get_version_run_image(oci_ref: &Reference) -> Result<u64> {
             .args(bon::vec![
                 "/bin/bash",
                 "-c",
-                "grep -Po '(?<=VERSION_ID=)\\d+' /usr/lib/os-release",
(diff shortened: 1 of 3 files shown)


process/drivers.rs
```diff
@@ -282,7 +282,7 @@ fn get_version_run_image(oci_ref: &Reference) -> Result<u64> {
             .args(bon::vec![
                 "/bin/bash",
                 "-c",
-                "grep -Po '(?<=VERSION_ID=)\\d+' /usr/lib/os-release",
+                "awk -F= '/^VERSION_ID=/ {gsub(\\"\\", \\"\\", $2); print $2}' /usr/lib/os-release",
```
gmpinder, line 285: You have syntax errors here. I'd suggest building locally to ensure it compiles

gmpinder (changes requested):

gmpinder (approved):

PR #298, opened 2024-12-14
docs: change file paths to match template

TheScreechingBagel:
the blue-build/template repo has these files ending in .yml not .yaml so probably should here as well


gmpinder (approved):

gmpinder:
Thanks for catching that!

Issue #362, opened 2025-02-10
Drivers do not support pre-release versions of tools


prydom:
Bluebuild v0.9.6 currently panics with the following error on a machine with podman `5.4.0-rc3` installed:

```
Could not determine strategy, need either docker version >=23, podman version >=4, or buildah version >=1.29 to continue
```

I believe this is because many semver comparison functions skip all pre-release versions.
Note that as of writing Fedora 41 Testing, 42, and rawhide all package a pre-release version of podman.

gmpinder:
This might require using `lenient_semver` for detecting the tool versions then as the `semver` crate doesn't support that. Luckily we're already making use of that for the image versions so that shouldn't be too difficult to figure out.

gmpinder:
I released `0.9.7` with a fix for this. Let me know if it works for you

prydom:
Unfortunately it didn't work.

Based on my reading of https://docs.rs/lenient_semver/latest/lenient_semver the crate doesn't provide `semver::VersionReq` which I think is what is at fault here. When a semver library compares a version range such as ">=4" I don't believe any version with a pre-release field will **ever** match.

This behavior is described [here](https://docs.rs/semver/1.0.25/src/semver/lib.rs.html#489) in a comment in the crate. 

>    Counterintuitively a `*` VersionReq does not match every possible version number. In particular, in order for *any* `VersionReq` to match a pre-release version, the `VersionReq` must contain at least one `Comparator` that has an explicit major, minor, and patch version identical to the pre-release being matched, and that has a nonempty pre-release component. Since `*` is not written with an explicit major,  minor, and patch version, and does not contain a nonempty pre-release component, it does not match any pre-release versions.

I've seen it also plainly described in the "Working With Prerelease Versions" section of https://masterminds.github.io/sprig/semver.html.

> According to the Semantic Version specification prereleases may not be API compliant with their release counterpart. It says, "A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version." SemVer comparisons using constraints without a prerelease comparator will skip prerelease versions. For example, >=1.2.3 will skip prereleases when looking at a list of releases while >=1.2.3-0 will evaluate and find prereleases.

Based on above we have two options, we could expand >=4 to the full >=4.0.0-0 or you could strip the pre-release field from versions before comparison.

```
~/my-ostree-build main*
❯ podman version -f json                                                                                             
[...]

PR #363, opened 2025-02-10
fix: Use lenient_semver for build drivers version check to handle pre…

gmpinder:
…-release versions


prydom:
Unfortunately, this didn't quite work. See https://github.com/blue-build/cli/issues/362#issuecomment-2649832019.

PR #365, opened 2025-02-14
chore: Retain cache for `rpm-ostree` & `dnf` post build

fiftydinar:
Just mimicking what Bluefin does here:

https://github.com/ublue-os/bluefin/blob/main/build_files/shared/clean-stage.sh

I'm not sure why they keep the cache from `rpm-ostree` & `dnf5`, but I just figured to make a PR, as it's easy to do.

### DIFF
--- a/scripts/post_build.sh
+++ b/scripts/post_build.sh
@@ -2,5 +2,7 @@
 
 set -euo pipefail
 
-rm -rf /tmp/* /var/*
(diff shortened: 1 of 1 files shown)


gmpinder:
They probably exclude those directories cause they have them mounted via cache mounts which already won't be in the image. I'll check later how the script is being called with the mounts we have

PR #388, opened 2025-03-13
feat(podman_driver): Invoke sudo when needed for privileged

dtomvan:
My motivation is that it's really a bad practice to just run `sudo bluebuild generate -o Containerfile`, as the file gets created as root and it's kind-of unsafe to the system too.

I'd rather have the tool invoke sudo by itselves for the things it needs it for, rather than wrapping the entire program in sudo.

This is especially the case for local (non-ci) builds.

I am not sure wether this is the perfect/correct code for this, but just looking at it this should work as long as that `cid` stuff doesn't also require root...


dtomvan:
Whoops, ran cargo check on the wrong branch. Fixed it.

gmpinder:
Yeah there's some issue with sending input from the user for typing the password with our logging system. I'm ok with calling it via `sudo`, but I think we'll have to make some additional adjustments to get this to work properly. We'll probably want to use `-A` where we can for `sudo` so that we can make use of graphical ways of taking password input, and if we can't use that, we have the command use `.status()` or `.output()` instead of our custom `.build_status()` function that captures output.

process/drivers/podman_driver.rs
```diff
@@ -528,8 +520,14 @@ impl RunDriver for PodmanDriver {
 }
 
 fn podman_run(opts: &RunOpts, cid_file: &Path) -> Command {
+    let argv0 = if opts.privileged && !nix::unistd::Uid::effective().is_root() {
+        "sudo"
+    } else {
+        "podman"
+    };
```
gmpinder, line 527:
```suggestion
    let use_sudo = opts.privileged && !nix::unistd::Uid::effective().is_root();
```

process/drivers/podman_driver.rs
```diff
@@ -528,8 +520,14 @@ impl RunDriver for PodmanDriver {
...
     let command = cmd!(
-        "podman",
+        argv0,
+        if argv0 != "podman" => "podman",
```
gmpinder, line 530:
```suggestion
        if use_sudo {
            "sudo"
        } else {
            "podman"
        },
        if use_sudo => [
            "-A",
            "--",
            "podman",
        ],
```

gmpinder (changes requested):
This could work for part of it.

process/drivers/podman_driver.rs
dtomvan: ah, you can see that I tried something like that before the fixup commit, and that didn't compile then, but apparently it works like this?

gmpinder: Yes, that's because the first parameter in that macro is what gets passed into `Command:: new` and every other is setting `.arg() `

dtomvan:
Uhh, I think you just merged main into here or something. Those are unrelated patches, right?

gmpinder:
> Uhh, I think you just merged main into here or something. Those are unrelated patches, right?

No these are intended. There were several things I had to change for this to work for all flows

gmpinder:
Thanks for bringing this to my attention!

PR #446, opened 2025-05-20
fix: pull akmods-extra only for bazzite (Fixes #441)

### LINKED ISSUE #441: Akmods build failure with Fedora 42
Found this as part of testing the nvidia driver installation via akmods.
Looks like Ublue only builds the `akmods-extra` images for the *bazzite* kernel ([see PR here](https://github.com/ublue-os/akmods/pull/311)).
When using other kernel (like base/main), the build fails if you try to use 42 as the base. 41 doesn't fail because there's a [3 months old image](https://github.com/ublue-os/akmods/pkgs/container/akmods-extra/363714353?tag=main-41), which is probably not very useful tbh.
`recipe/src/module.rs` should probably be updated to reflect this and avoid pulling that image when not using *bazzite* as the base.

franute:
When using fedora 42 with the main images, the build will fail due to the non-existing akmods-extra images.
It looks like, these images will only exist for the `bazzite` based ones.

### DIFF
--- a/recipe/src/akmods_info.rs
+++ b/recipe/src/akmods_info.rs
@@ -3,7 +3,7 @@ use bon::Builder;
 #[derive(Debug, Clone, Builder, PartialEq, Eq, Hash)]
 pub struct AkmodsInfo {
     #[builder(into)]
-    pub images: (String, String, Option<String>),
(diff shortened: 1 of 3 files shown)


gmpinder (approved):
Awesome! Thanks for the changes

PR #578, opened 2025-10-18
feat(labels): Custom Image Labels

Zyell:
This is the feature as discussed in discord: https://discord.com/channels/1198292321037455360/1428080454417776822

The goal is to allow custom labels to be declared in the recipe and applied to the image like the following:

```yaml
labels:
    org.container.name: name
    org.test.label: test
```


process/drivers/traits.rs
```diff
@@ -479,21 +479,7 @@ pub trait RechunkDriver: RunDriver + BuildDriver + ContainerMountDriver {
...
-            ),
-            format_args!("org.opencontainers.image.title={}", &opts.name),
-            format_args!("org.opencontainers.image.description={}", &opts.description),
-            format_args!("org.opencontainers.image.source={}", &opts.repo),
-            format_args!("org.opencontainers.image.base.digest={}", &opts.base_digest),
-            format_args!("org.opencontainers.image.base.name={}", &opts.base_image),
-            "org.opencontainers.image.created=<timestamp>",
-            "io.artifacthub.package.readme-url=https://raw.githubusercontent.com/blue-build/cli/main/README.md",
-        );
+        let labels = generate_labels(opts.recipe_path)?;
```
gmpinder: The `generate_labels` function appears to include the `LABEL` instruction. This rechunk function requires the labels to be `\n` delimited without that `LABEL` instruction.

process/labels.rs
```diff
@@ -0,0 +1,180 @@
...
+///
+/// # Errors
+///
+/// Returns an error if:
+/// - The recipe file cannot be parsed
+/// - Unable to retrieve repository URL
+/// - Unable to get metadata for the base image
+/// - Unable to generate the base image reference
+#[cached(result = true, key = "PathBuf", convert = r"{ recipe_path.into() }")]
+pub fn generate_labels(recipe_path: &Path) -> miette::Result<String> {
```
gmpinder: These functions would probably do better as a method on `Recipe` instead.

process/drivers/traits.rs
Zyell: Ah I missed this!  I will correct it, thanks!

template/src/lib.rs
```diff
@@ -36,13 +35,14 @@ pub struct ContainerFileTemplate<'a> {
...
     build_scripts_dir: &'a Path,
-    repo: &'a str,
     base_digest: &'a str,
     nushell_version: Option<&'a MaybeVersion>,
 
     #[builder(default)]
     build_features: &'a [String],
     build_engine: BuildEngine,
+
+    labels: &'a str,
```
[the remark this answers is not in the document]
Zyell: Converting this to a HashMap causes issues with Template derivation:

```
the method `askama_auto_escape` exists for reference `&&AutoEscaper<'_, &HashMap<String, String>, Text>`, but its trait bounds were not satisfied [E0599]
Note: the following trait bounds were not satisfied:
`&HashMap<String, String>: std::fmt::Display`
which is required by `&&AutoEscaper<'_, &HashMap<String, String>, Text>: AutoEscape`
```
I could implement a newtype around the HashMap to implement Display and allow easy conversion for the template.  Or is there a different approach you would recommend?

process/labels.rs
Zyell: Oh nevermind,  I  need access to both the recipe and the Driver to generate the default labels, in particular because of this:
```rust
  let image_metada = Driver::get_metadata(
      GetMetadataOpts::builder()
          .image(&recipe.base_image_ref()?)
          .build(),
  )?;
  let base_digest = image_metada.digest();
```
I will need to move this to the generate file instead and then passing in the full key value mapping to the rechunk options fixes the dependency mixing.

gmpinder: Which is why I suggested the root crate as that's the one that ties everything together. Everything in `process/` is the `blue-build-process-management` crate and everything in `src/` is the `blue-build` crate which is basically the orchestrator portion of the code base and what creates the final bin.

gmpinder:
> > Thanks for taking the time to contribute! My comments are mostly around trying to better compartmentalize where code should be so that we remove the dependency on `blue-build-recipe` in the `blue-build-process-management` crate.
> 
> I have pushed the changes I have so far, but I do have one issue I haven't been able to solve yet, though it causes no actual problem. In the Containerfile.j2 this line is getting extra new lines so that the generated Containerfile has an extra line between each label.
> 
> ```
> {% for (name, value) in labels %}
>    LABEL {{ name }}="{{ value }}"
> {% endfor %}
> ```
> 
> I will come back to this one later, it is odd.
> 
> I want to clean up the merging of the custom and default mappings a little, but I ran out of time to do it today. I will follow up tomorrow.
> 
> In the meantime let me know if this structurally solves your concerns and I will wrap up these last items tomorrow. Thanks!

You can add a `-` to the opening bracket like so:

```jinja
{%- for (name, value) in labels %}
LABEL {{ name }}="{{ value }}"
{%- endfor %}
```

Make sure you don't have indents on the templated portion of the file as it will show up indented on the final output.

process/drivers/traits.rs
```diff
@@ -479,21 +478,12 @@ pub trait RechunkDriver: RunDriver + BuildDriver + ContainerMountDriver {
...
-            format_args!("org.opencontainers.image.title={}", &opts.name),
-            format_args!("org.opencontainers.image.description={}", &opts.description),
-            format_args!("org.opencontainers.image.source={}", &opts.repo),
-            format_args!("org.opencontainers.image.base.digest={}", &opts.base_digest),
-            format_args!("org.opencontainers.image.base.name={}", &opts.base_image),
-            "org.opencontainers.image.created=<timestamp>",
-            "io.artifacthub.package.readme-url=https://raw.githubusercontent.com/blue-build/cli/main/README.md",
-        );
+        let labels = opts.labels;
+        let label_string = labels
```
gmpinder: Should be able to just use the labels:


```suggestion
        let label_string = opts.labels
```

recipe/src/recipe.rs
```diff
@@ -166,4 +172,148 @@ impl Recipe<'_> {
             .into_iter()
             .collect()
     }
+
+    #[must_use]
+    pub fn generate_labels(
+        &self,
+        default_labels: &BTreeMap<String, String>,
+    ) -> BTreeMap<String, String> {
+        aggregate_labels(default_labels, &self.labels.clone().unwrap_or_default())
```
gmpinder: Might as well just have the contents of `aggregate_labels` moved in here.

template/templates/Containerfile.j2
```diff
@@ -51,11 +51,6 @@ RUN \
...
-LABEL org.opencontainers.image.title="{{ recipe.name }}"
-LABEL org.opencontainers.image.description="{{ recipe.description }}"
-LABEL org.opencontainers.image.source="{{ repo }}"
-LABEL org.opencontainers.image.base.digest="{{ base_digest }}"
-LABEL org.opencontainers.image.base.name="{{ recipe.base_image }}:{{ recipe.image_version }}"
-LABEL org.opencontainers.image.created="{{ self::current_timestamp() }}"
-LABEL io.artifacthub.package.readme-url=https://raw.githubusercontent.com/blue-build/cli/main/README.md
+{% for (name, value) in labels %}
+   LABEL {{ name }}="{{ value }}"
+{% endfor %}
```
gmpinder: Like how I mentioned in my other comment:


```suggestion
{%- for (name, value) in labels %}
LABEL {{ name }}="{{ value }}"
{%- endfor %}
```

gmpinder (changes requested):
This is looking great! Just some nits

recipe/src/recipe.rs
Zyell: Yes, very true.  I won't have the same testing issue I had with the prior structure.

PR #602, opened 2025-11-09
feat(iso): Add --web-ui option

gmpinder:
Version [v1.4.0](https://github.com/JasonN3/build-container-installer/releases/tag/v1.4.0) has support for building an ISO using the new WebUI for Anaconda. We also pin the version for consistency sake.


john-holt4:
Just want to put this bug out here https://github.com/blue-build/cli/issues/605

PR #610, opened 2025-11-17
feat: support rechunking with build-chunked-oci

HastD:
Add support for rechunking images with [`rpm-ostree compose build-chunked-oci`](https://coreos.github.io/rpm-ostree/build-chunked-oci/). This is available in the CLI with the `--build-chunked-oci` option, with an optional `--max-layers` argument to specify the maximum number of layers in the rechunked image.


gmpinder:
Ok that change seemed to work. I'm going to update your branch real quick.

PR #614, opened 2025-11-21
chore(exports.sh): Export `OS_ARCH` variable to be from the container…

fiftydinar:
… image instead from the runner

It also sets `amd64` to `x86_64` and `arm64` to `aarch64`, as I think it's more standard this way.


gmpinder (approved):

gmpinder:
There seems to be some issue with the `podman manifest create` command that's affecting all builds. I'll work on that separately from this change.

PR #616, opened 2025-11-21
chore(build_chunked_oci): Clean up and consolidate chunked logic

gmpinder:


process/drivers/opts/build_chunked_oci.rs
```diff
@@ -1,25 +1,44 @@
...
 #[builder(derive(Debug, Clone))]
-#[non_exhaustive]
 pub struct BuildRechunkTagPushOpts<'scope> {
     pub build_tag_push_opts: BuildTagPushOpts<'scope>,
     pub rechunk_opts: BuildChunkedOciOpts,
 }
 
 #[derive(Debug, Clone, Copy, Builder)]
 #[builder(derive(Debug, Clone))]
-#[non_exhaustive]
```
HastD: What's the reason for removing `#[non_exhaustive]`? I added it under the reasoning that build-chunked-oci may add more options in the future; `#[non_exhaustive]` ensures that, if that happens, it won't be a breaking change to add more fields to `BuildChunkedOciOpts`.

process/drivers/opts/build_chunked_oci.rs
```diff
@@ -1,25 +1,44 @@
...
+    pub format_version: BuildChunkedOciFormatVersion,
 
     /// Maximum number of layers to use. Currently defaults to 64 if not specified.
-    pub max_layers: Option<NonZero<u32>>,
+    #[builder(default = DEFAULT_MAX_LAYERS)]
+    pub max_layers: NonZeroU32,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum BuildChunkedOciFormatVersion {
```
HastD: This should also be `#[non_exhaustive]` to account for the possibility of additional format versions in the future.

HastD (commented):
LGTM aside from the below minor nitpicks.

process/drivers/opts/build_chunked_oci.rs
gmpinder: This is mostly because these are supposed to be APIs meant for the CLI and not necessarily for other applications. I like to get compiler errors whenever I add something new so that I can go and make the changes in all the places that use it.

PR #653, opened 2025-12-23
feat(stages): Add the ability to set the platform for a stage

gmpinder:
This allows setting a specific platform for a stage. This can be helpful if an image only has one available platform, and you want to make sure the build uses it. Another use is cross-compiling software. Often it can be easier to target another platform when you compile a program since it's generally faster to run that cross-compilation on native hardware than it is to emulate the platform and build as if it's "native".


gmpinder:
Integration test will fail until https://github.com/blue-build/schema/pull/14 is merged.

PR #663, opened 2026-01-06
feat: add option to remove base image after building new image

HastD:
The `--remove-base-image` option removes the base image after building, but before build-chunked-oci runs, freeing additional disk space. This should be useful on runners with relatively little disk space, where the high space requirements of build-chunked-oci can be problematic.

### DIFF
--- a/process/drivers/buildah_driver.rs
+++ b/process/drivers/buildah_driver.rs
@@ -1,5 +1,5 @@
 use blue_build_utils::{
-    container::ContainerId, credentials::Credentials, secret::SecretArgs, semver::Version,
+    container::ContainerId, credentials::Credentials, secret::SecretArgs, semver::Version, sudo_cmd,
 };
 use colored::Colorize;
 use comlexr::{cmd, pipe};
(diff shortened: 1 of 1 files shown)


HastD:
Good, looks like CI is passing now.

HastD:
Something is failing in some of the tests, but it's hard to see exactly what from the info it prints; I've made a small change to the `RunDriver` implementations so the relevant methods show stderr of the failed command in case of failure.

process/drivers/buildah_driver.rs
```diff
@@ -329,3 +330,63 @@ impl BuildDriver for BuildahDriver {
         Ok(())
     }
 }
+
+impl ImageStorageDriver for BuildahDriver {
+    fn remove_image(opts: super::opts::RemoveImageOpts) -> Result<()> {
+        trace!("BuildahDriver::remove_image({opts:?})");
+
+        let output = {
+            let c = cmd!("buildah", "rmi", opts.image.to_string());
```
gmpinder: This should use `sudo_cmd!` like the podman driver does for priveleged

HastD: Done, and also done for the methods in the `BuildDriver` implementation where the podman driver uses `sudo_cmd!`.

gmpinder: Boy am I glad that you're catching things that I've missed

ruslanchema12-beep:
Stripe.com

PR #716, opened 2026-02-27
feat(mounts): Implement mounts option for Recipe and Module

Aex12:
This PR implements the `mounts` option as proposed in https://github.com/blue-build/schema/pull/16

### DIFF
--- a/recipe/src/recipe.rs
+++ b/recipe/src/recipe.rs
@@ -92,7 +92,7 @@ pub struct Recipe {
     pub labels: Option<HashMap<String, String>>,
 
     /// The mounts to add to the image.
-    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(skip_serializing_if = "Vec::is_empty", default)]
     #[builder(default)]
     pub mounts: Vec<Mount>,
 }
--- a/recipe/src/recipe.rs
+++ b/recipe/src/recipe.rs
@@ -15,7 +15,7 @@ use miette::{Context, IntoDiagnostic, Result};
 use oci_client::Reference;
 use serde::{Deserialize, Serialize};
 
-use crate::{Module, ModuleExt, StagesExt, maybe_version::MaybeVersion};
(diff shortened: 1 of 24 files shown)


recipe/src/recipe.rs
```diff
@@ -90,6 +90,10 @@ pub struct Recipe {
     /// This hashmap provides custom labels from ther use to the image
     #[serde(skip_serializing_if = "Option::is_none")]
     pub labels: Option<HashMap<String, String>>,
+
+    /// The mounts to add to the image.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub mounts: Option<Vec<Mount>>,
```
gmpinder:
```suggestion
    #[builder(into, default)]
    #[serde(skip_serializing_if = "Vec::is_empty", default)]
    pub mounts: Vec<Mount>,
```

I have some refactoring to do with other `Option<Vec>` in this file. It makes it easier to iterate over the values since an empty vec is essentially `None`. I'm doing the same thing for [secrets](https://github.com/blue-build/cli/blob/857442c22cda5ad0f9db017f20dc3c40007a1129/recipe/src/module.rs#L40).

gmpinder (changes requested):

gmpinder (changes requested):
Looking good so far can you add some of this new functionality to the recipe in `./integration-test/test-repo/`? That one is one that gets built in the integration tests. The one you were updating, which was needed, is for testing the schema.

gmpinder:
I also left a comment in discord about updating the constant for the schema base URL. Doing that will pull the new schema changes before they get fully merged in.

gmpinder (changes requested):

gmpinder:
Sorry for not getting back to you, but it looks like the `sharing` option does exist, but [only supports `shared` and `locked`](https://github.com/containers/buildah/blob/d724f1deba480b615e4ccd0c1affb3daa80ca226/internal/volumes/volumes.go#L607-L635), so maybe we just support those values for now.

PR #749, opened 2026-04-10
feat: Add environment variable expansion in top-level recipe fields

gmpinder:
This adds the ability to set environment variables to control the top-level fields of the recipe. This is useful for easily changing where you pull your images from or bumping the base image version without updating the file directly. It allows matrixing in Gitlab CI or Github Actions to be more customizable.

This follows similar POSIX shell expansion logic. Information about the implementation can be found [here](https://docs.rs/shellexpand/latest/shellexpand/)

### DIFF
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -15,6 +15,7 @@ constcat = "0.6.1"
 directories = "6.0.0"
 docker_credential = "1.3.2"
 process_control = { version = "4.2.2", features = ["crossbeam-channel"] }
+shellexpand = "3.1.2"
 
 bon.workspace = true
 cached.workspace = true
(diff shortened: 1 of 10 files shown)


utils/Cargo.toml
```diff
@@ -15,6 +15,7 @@ constcat = "0.6.1"
 directories = "6.0.0"
 docker_credential = "1.3.2"
 process_control = { version = "4.2.2", features = ["crossbeam-channel"] }
+shellexpand = "3.1.2"
```
HastD: Since we're only using environment variable expansion (not tilde-expansion), we can get rid of an unnecessary dependency by disabling [some default-enabled features](https://docs.rs/crate/shellexpand/latest/features):
```toml
shellexpand = { version = "3.1.2", default-features = false, features = ["base-0"] }
```

HastD (changes requested):

utils/Cargo.toml
gmpinder: That's a good idea

HastD (approved):

PR #768, opened 2026-04-29
fix: --pull=always, not --pull=true

HastD:
According to the documentation for `podman build`/`buildah build`, the valid options for `--pull` are "always", "missing", "never", and "newer"; "true" isn't a valid option, so I'm not sure what it does but I think "always" is what's intended here.


gmpinder (approved):
You know, you would think that it would error out if this was wrong. But, okay.

PR #789, opened 2026-05-31
feat(recipe): Recipe V2

### LINKED ISSUE #180: arriving at a stable, functional, and sensical recipe standard
(this issue is filed under CLI, because this is the only place changes would have to be made to implement changes to the recipe format)

_I had an epiphany._ But let me preface a bit; if the purpose of the _recipe_ is to be a versatile standard that describes the build and publish process of a native container image, is it as good as that as can be? And if not, aren't we in the _broad_ ideal moment of time to evolve it?

Here's how it _could_ look:
```yaml
base-image:
    url: ghcr.io/ublue-os/silverblue-main
    tag: 40
    public-key: https://... # unimplemented, needed for base image verification, would be optional, and automatically supplied for ublue etc.
    type: ostree # unimplemented, might be needed for effectively supporting different base images, which might need things like running `ostree container commit` 
result-image:
    name: weird-os
    description: This is my personal OS image.
    tags: # replaces the "latest" and timestamp tags
       - gts
stages:
    ...
modules:
    ...
```

That's a sketch, but IMO it's already better than what we have now. Ideas and discussion appreciated here, though.

Multiple versions of the recipe standard could be supported in a way similar to other projects, say, [`docker-compose.yml`](https://stackoverflow.com/questions/76156527/what-does-the-first-line-in-the-docker-compose-yml-file-that-specifies-the-ve) do. I'd opt for using the `$schema:` key to specify the version; that would force us to supply jsonschema for all supported versions, and the user to update both the autocomplete source and the actual version in their recipes at the same time. This would set a solid foundation for breaking recipe changes in the future. If rust can do jsonchema validation, maybe we could even get multi-version support in CLI with not much extra effort?

gmpinder:
The implementation for Recipe V2. Schema changes are in https://github.com/blue-build/schema/pull/19.

Closes #180

### DIFF
--- a/utils/src/constants.rs
+++ b/utils/src/constants.rs
@@ -104,6 +104,7 @@ pub const SUDO_ASKPASS: &str = "SUDO_ASKPASS";
 // Misc
 pub const BLUE_BUILD: &str = "bluebuild";
 pub const BUILD_SCRIPTS_IMAGE_REF: &str = "ghcr.io/blue-build/cli/build-scripts";
+pub const BLUE_BUILD_DEFAULT_IMAGE: &str = "ghcr.io/blue-build/base-images/fedora-base:latest";
 pub const BLUE_BUILD_IMAGE_REF: &str = "ghcr.io/blue-build/cli";
 pub const BLUE_BUILD_MODULE_IMAGE_REF: &str = "ghcr.io/blue-build/modules";
 pub const BLUE_BUILD_SCRIPTS_DIR_IGNORE: &str = "/.bluebuild-scripts_*";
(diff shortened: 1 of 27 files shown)


xynydev (dismissed):
I am not qualified to review Rust code. If it works, it works. I would assume that it works. As long as it parses recipe v1 by default and v2 if `version: 2` is set, we're good.

utils/src/constants.rs
```diff
@@ -125,7 +126,8 @@ pub const GITHUB_CHAR_LIMIT: usize = 8100; // Magic number accepted by Github
 pub const DEFAULT_MAX_LAYERS: NonZeroU32 = NonZeroU32::new(128).unwrap();
 
 // Schema
-pub const SCHEMA_BASE_URL: &str = "https://schema.blue-build.org";
+// pub const SCHEMA_BASE_URL: &str = "https://schema.blue-build.org";
+pub const SCHEMA_BASE_URL: &str = "https://recipe-v2.schema-e29.pages.dev";
```
gmpinder:
```suggestion
pub const SCHEMA_BASE_URL: &str = "https://schema.blue-build.org";
```

PR #790, opened 2026-06-01
feat: support rechunking using Chunkah

HastD:
Add a `--chunkah` option to `bluebuild build` that rechunks the image using [Chunkah](https://github.com/coreos/chunkah), a more modern, content-agnostic rechunker that's actively maintained and should offer various improvements over build-chunked-oci and the legacy rechunker.

This is implemented using a fairly generic "post-build hook" mechanism that allows a post-processing step (such as rechunking) to be applied after building the image but before tagging and pushing.


process/drivers/skopeo_driver.rs
```diff
@@ -12,37 +13,46 @@ impl super::OciCopy for SkopeoDriver {
...
-                    "skopeo"
-                },
-                if use_sudo && blue_build_utils::has_env_var(blue_build_utils::constants::SUDO_ASKPASS) => [
-                    "-A",
-                    "-p",
+        let mut initial_args = Vec::<OsString>::new();
+        if use_sudo {
+            initial_args.push("sudo".into());
+            if blue_build_utils::has_env_var(blue_build_utils::constants::SUDO_ASKPASS) {
+                initial_args.push("-A".into());
```
gmpinder: I think I should work on that `cmd_append!` macro. This feels more messy than it needs to be.

process/drivers/opts/image_storage.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub image: &'scope str,
+
+    /// Path to write output, or `None` if output should be returned to the caller.
+    pub output_path: Option<&'scope Path>,
+}
+
+#[derive(Debug, Clone, Copy, Builder)]
+#[builder(derive(Debug, Clone))]
+pub struct RemoveImageOpts<'scope> {
+    pub image: &'scope str,
```
gmpinder: Could we try to keep this as `&Reference`? I'm noticing extra `.to_string()` calls being added which means more allocations. Also it helps to keep the API stricter.

gmpinder (changes requested):

process/drivers/skopeo_driver.rs
gmpinder: Ok, went ahead and made it. It's called `cmd_mut!`. Released on version 1.5.0.

https://gitlab.com/wunker-bunker/comlexr/-/commit/1523e3e65c7d344c1ec1874b5f6567ce84f134f9

HastD: Oh nice, that will be very useful. I'll go ahead and modify the code to use it.

HastD: Though actually wait, this might not *quite* solve the messiness here, since the issue is with needing to dynamically determine the command name (not subsequent arguments).

process/drivers/opts/image_storage.rs
HastD: Yeah, I can change it back.

HastD: Though also I wouldn't worry too much about allocations; I'm pretty sure the vast majority of BlueBuild's runtime is spent on I/O done by external processes. I haven't seen profiling data, but I'd be surprised if allocations account for more than 0.1% of runtime.

process/drivers/skopeo_driver.rs
gmpinder: You can use a block for the initial command name. 

```rust
let command = cmd!(
{
  if this {
    "that"
  } else {
    "other"
  }
},
"arg1",
"arg2",
);
```

process/drivers/opts/image_storage.rs
gmpinder: > Though also I wouldn't worry too much about allocations; I'm pretty sure the vast majority of BlueBuild's runtime is spent on I/O done by external processes. I haven't seen profiling data, but I'd be surprised if allocations account for more than 0.1% of runtime.

That's a fair point, but the extra safety in the API is still useful.

process/drivers/skopeo_driver.rs
HastD: Right but it's still awkward to do something like "optionally insert sudo at the beginning of the command", since it makes the subsequent argument conditional, too. What I'd ideally want is something like
```rs
cmd!(
    if condition => "sudo",
    "foo",
    // ...
)
```
where it automatically handles making the first argument where the condition is met into the command name. The new `cmd_mut!` macro is nice, but I don't think it will actually simplify things here; the problem is that depending on a combination of conditions, any of `sudo`, `podman`, or `skopeo` could be the command name, and where the argument list picks up after that also depends on those conditions.

gmpinder: We have the `sudo_cmd!` macro in the project. That does that for you

gmpinder: https://github.com/blue-build/cli/blob/main/utils%2Fsrc%2Fmacros.rs#L84

process/drivers/buildah_driver.rs
```diff
@@ -343,6 +345,25 @@ impl BuildDriver for BuildahDriver {
...
+            .stdout(stdout)
+            .output()
+            .into_diagnostic()?;
+
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            bail!("Failed to inspect image {}:\n{}", opts.image, stderr);
+        }
+
+        Ok(opts.output_path.map(|_| output.stdout))
```
gmpinder: I'm a little confused, if the output path is some, then wouldn't we want that to go to the file and not return from the function?

gmpinder:
The cargo test job appears to be failing from some compilation error with OCI client

process/drivers/buildah_driver.rs
HastD: You're right, got the condition backwards there. Fixing it.

gmpinder:
I'm seeing an error for the container chunkah build:

```
[21:25:50 TRACE blue_build_process_management::drivers::skopeo_driver:15] => SkopeoDriver::copy_oci(CopyOciOpts { src_ref: OciDir("/tmp/bluebuild-tmp-izCTpT3KHP/oci-out"), dest_ref: LocalStorage(Reference { registry: "localhost", mirror_registry: None, repository: "cli/test-chunkah", tag: Some("latest_linux_amd64"), digest: None }), privileged: false, retry_count: 0, podman_unshare: true })
[21:25:50 TRACE blue_build_process_management::drivers::skopeo_driver:29] => "podman" "unshare" "skopeo" "copy" "--all" "--" "oci:/tmp/bluebuild-tmp-izCTpT3KHP/oci-out" "containers-storage:localhost/cli/test-chunkah:latest_linux_amd64"
[21:25:50 containers-storage:localhost/cli/test-chunkah] => Error: please use unshare with rootless
[21:25:50 ERROR blue_build::commands:33] => Failed:
  × Failed to copy oci:/tmp/bluebuild-tmp-izCTpT3KHP/oci-out to containers-storage:localhost/cli/test-chunkah:latest_linux_amd64
```

HastD:
@gmpinder I think it should be fixed now, but CI is failing due to network errors so I'm not 100% sure.

gmpinder (approved):
Just waiting on some builds to pass

Issue #800, opened 2026-06-11
Failed to sign Image with keyless Signing in Gitlab CI


MaxHerrmannSVA:
Greetings,

when building an Image in a Gitlab Pipeline I get an error as we tried to sign the images without a cosign.key file

```text
<gitlab-server>:br-feat_keyless-signing-latest-43 =>  > [stage-4  1/13] RUN --mount=type=bind,from=stage-keys,src=/keys,dst=/tmp/keys   mkdir -p /etc/pki/containers/   && cp /tmp/keys/* /etc/pki/containers/:
<gitlab-server>:br-feat_keyless-signing-latest-43 => ------
<gitlab-server>:br-feat_keyless-signing-latest-43 => Containerfile.ygA-7NRyepY:39
<gitlab-server>:br-feat_keyless-signing-latest-43 => --------------------
<gitlab-server>:br-feat_keyless-signing-latest-43 =>   38 |     # Key RUN
<gitlab-server>:br-feat_keyless-signing-latest-43 =>   39 | >>> RUN --mount=type=bind,from=stage-keys,src=/keys,dst=/tmp/keys \
<gitlab-server>:br-feat_keyless-signing-latest-43 =>   40 | >>>   mkdir -p /etc/pki/containers/ \
<gitlab-server>:br-feat_keyless-signing-latest-43 =>   41 | >>>   && cp /tmp/keys/* /etc/pki/containers/
<gitlab-server>:br-feat_keyless-signing-latest-43 =>   42 |     # Bin RUN
<gitlab-server>:br-feat_keyless-signing-latest-43 => --------------------
<gitlab-server>:br-feat_keyless-signing-latest-43 => ERROR: failed to build: failed to solve: process "/bin/sh -c mkdir -p /etc/pki/containers/   && cp /tmp/keys/* /etc/pki/containers/" did not complete successfully: open /keys: no such file or directory
```

But I think I already found the bug for this issue

When looking into https://github.com/blue-build/cli/blob/main/template/templates/stages.j2#L47 it is checked if we have a cosign.pub file present

This is not the case here: https://github.com/blue-build/cli/blob/main/template/templates/Containerfile.j2#L31
So if Im correct we only have to add the if condition in the Containerfile.j2 and the build can continue

PR #801, opened 2026-06-11
Fix keylessImage signing

MaxHerrmannSVA:
See #800 for the details :)

Thank you


gmpinder (approved):
Nice catch! Thanks for the fix

PR #806, opened 2026-06-30
feat(cosign): allow configurable sign/verify args via env vars

jtsylve:
The cosign signing driver builds fixed `cosign sign` / `cosign verify` commands with no way to pass extra flags. cosign defaults `--tlog-upload=true` on sign (publishing image digests to the public Rekor log and requiring internet) and, symmetrically, `cosign verify` requires a tlog entry by default — so an image signed offline cannot be verified by the build's own sign-and-verify step.

## Change

Add two **opt-in** env vars whose whitespace-separated tokens are appended to the respective command:

- `BB_COSIGN_SIGN_ARGS` → appended to `cosign sign`
- `BB_COSIGN_VERIFY_ARGS` → appended to `cosign verify`

Both are unset by default, so there is **no behavior change** for existing users. For a private/offline registry:

```
BB_COSIGN_SIGN_ARGS="--tlog-upload=false"
BB_COSIGN_VERIFY_ARGS="--insecure-ignore-tlog=true"
```

## Notes

- Env-var names are declared in `utils/src/constants.rs` per the existing convention and documented with rustdoc.
- Token splitting is factored into a small `split_extra_args` helper with a unit test (whitespace-only → no args, single flag, multiple flags). Shell quoting is intentionally not supported; each flag must be a single whitespace-free token.
- `cargo fmt --check`, `cargo clippy`, and the new unit test pass.

### DIFF
--- a/process/drivers/cosign_driver.rs
+++ b/process/drivers/cosign_driver.rs
@@ -1,7 +1,9 @@
 use std::{fmt::Debug, fs, path::Path};
 
 use blue_build_utils::{
-    constants::{COSIGN_PASSWORD, COSIGN_PUB_PATH, COSIGN_YES},
+    constants::{
+        BB_COSIGN_SIGN_ARGS, BB_COSIGN_VERIFY_ARGS, COSIGN_PASSWORD, COSIGN_PUB_PATH, COSIGN_YES,
+    },
     credentials::Credentials,
     semver::Version,
 };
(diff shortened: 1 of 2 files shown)


gmpinder (changes requested):
I feel like this is something that should be an argument to the CLI itself, especially if we want this to also work with the [`sigstore`](https://crates.io/crates/sigstore) driver too as the cosign driver isn't the only thing that is used for signing and verification. As a matter of fact, the cosign driver will be removed at a later point in favor of the native rust code.

PR #807, opened 2026-06-30
feat(oci-client): use plain HTTP for loopback registries

jtsylve:
`OciClientDriver::get_metadata` builds its oci-client with `ClientConfig::default()`, whose protocol is HTTPS. The CLI also links reqwest with `rustls-tls` (webpki roots) and hardcodes `accept_invalid_certificates: false`, so a self-signed TLS registry is not trusted either. The result: `bluebuild build` cannot inspect an image on a plain-HTTP local registry — the post-push manifest fetch fails with a connect error against `https://localhost:5000/...`.

## Change

Treat loopback registries (`localhost`, `127.0.0.0/8`, `::1` — any port) as plain HTTP, exactly as cosign, containerd, and docker already do. Everything else stays HTTPS, so there is **no behavior change** for real registries and **no new configuration** to set.

## Notes

- Loopback detection is factored into an `is_loopback_registry` helper that handles an optional `:port` suffix and bracketed IPv6 literals, with unit tests covering loopback (incl. `127.0.0.0/8`, `::1`, `[::1]:5000`) and remote (incl. `2001:db8::1`, `localhost.example.com`) cases.
- `oci-client` has no auto-negotiating protocol (only `Http` / `Https` / `HttpsExcept`). The client is built per-registry and only ever contacts the already-confirmed-loopback registry, so `ClientProtocol::Http` is used directly.
- `cargo fmt --check`, `cargo clippy`, and the new unit tests pass.

### DIFF
--- a/process/drivers/oci_client_driver.rs
+++ b/process/drivers/oci_client_driver.rs
@@ -17,7 +17,18 @@ impl InspectDriver for OciClientDriver {
     fn get_metadata(opts: GetMetadataOpts) -> Result<ImageMetadata> {
         #[cached(result = true, key = "String", convert = r"{image.to_string()}")]
         fn inner(image: &Reference) -> Result<ImageMetadata> {
-            let client = oci_client::Client::new(ClientConfig::default());
+            // Speak plain HTTP to loopback registries (localhost, 127.0.0.0/8,
+            // ::1 — any port), matching the convention of cosign, containerd,
+            // and docker. Everything else stays HTTPS. No configuration needed.
+            let protocol = if is_loopback_registry(image.registry()) {
+                oci_client::client::ClientProtocol::Http
+            } else {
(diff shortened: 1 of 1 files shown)


process/drivers/oci_client_driver.rs
```diff
@@ -81,3 +92,66 @@ impl InspectDriver for OciClientDriver {
...
+        || host
+            .parse::<std::net::IpAddr>()
+            .is_ok_and(|ip| ip.is_loopback())
+}
+
+#[cfg(test)]
+mod test {
+    use super::is_loopback_registry;
+
+    #[test]
```
gmpinder, line 120: Try making use of `rstest` instead

https://crates.io/crates/rstest

gmpinder (changes requested):

PR #809, opened 2026-07-02
feat: pull with retries before building

HastD:
This should reduce the frequency of build failures caused by network/server issues when pulling a remote image before building.

Additionally, improve the retry utility functions:
* Make the retry functions accept a `Duration` rather than a `u64` representing seconds. This makes it clear at the call site that the argument represents a duration.
* Make `retry_async` use `tokio::time::sleep` instead of the blocking `std::thread::sleep`. Async functions shouldn't block for an extended period of time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -578,6 +578,7 @@ dependencies = [
  "shellexpand",
  "syntect",
  "tempfile",
+ "tokio",
  "uuid",
  "which",
  "yaml_serde",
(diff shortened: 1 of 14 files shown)


gmpinder (approved):

gmpinder (changes requested):
Looks like there's a bug in the logic you introduced. It's trying to pull images that don't exist yet

HastD:
Oops, I see the issue, it's trying to pull the image to be created, but it should pull the *base* image instead. One sec, fixing it.

gmpinder (approved):

PR #811, opened 2026-07-08
ci: faster free-disk-space action; update empty-files-repo

HastD:
* Use `hastd/free-disk-space` action instead of `ublue-os/remove-unwanted-software` to free disk space in CI. This is already used in `blue-build/github-action`; it's faster and can be configured to automatically skip running if the runner has enough disk space already (which is often the case as runners seem to be instantiated with more disk space a lot of the time now).
* Update base image in `empty-files-repo` to version 44, as version 42 is no longer being built and was producing an error.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,7 +96,9 @@ jobs:
 
     steps:
       - name: Maximize build space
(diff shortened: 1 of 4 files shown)


gmpinder (approved):
Thanks for fixing something I've been lazy about 😅

PR #832, opened 2026-07-26
feat: prefer local metadata, making local builds easier

FlorianLudwig:
The metadata for the `base-image` is fetched from the registry, even if the image is available locally.

This PR instead prefers the local available metadata and only falls back to the registry if image is not available locally or parsing local data fails.


gmpinder:
I appreciate the attempt to make operations faster. However, this change goes against the general architecture of the driver system and we will be refactoring the code base in preparation for v1.

PR #870, opened 2026-08-30
chore: Migrate to Renovate for automated dependency updates

jonerrr:
Migrates dependency updates from Dependabot to Renovate.

This also adds a custom manager for the pinned runtime images referenced in Rust:

- Cosign
- Build Container Installer
- Rechunk

Their versions are tracked through GitHub Releases so Renovate can apply the inherited release-age policy.

Should we also consider pinning the remaining `latest` image references by digest and letting Renovate update those digests?


gmpinder:
> Should we also consider pinning the remaining latest image references by digest and letting Renovate update those digests?

Yes that would be a good idea. There are also image references in the `Earthfile`s that should be updated.

jonerrr:
> Yes that would be a good idea. There are also image references in the `Earthfile`s that should be updated.

Just implemented both of those changes. One thing to note is that renovate will make a PR to update the `fedora-toolbox` image to the latest version of fedora, but you can just ignore it or limit the version number in the renovate config. Also if the PR "spam" is too much, we can consider grouping some updates into a single PR.

gmpinder:
@jonerrr mind rebasing your changes? I just merged in support for forks. I want to make sure that everything builds properly.

gmpinder (approved):

